PR #6, opened 2017-12-19
Jw-1197 Possible fix: Add waiting in Disable-DockerDriver

krdln:
Should make `Disable-DockerDriver` more robust.

This is a possible fix for JW-1197 (docker driver was not enabled (problems with //.pipe/Contrail), ~~~it should also fix the issue when docker driver was not enabled due to `HNS failed with error : Unspecified error`.~~~

~~~_In theory_, this should make retrying `Enable-DockerDriver` unnecessary (d2c9b21c74a1eba3afb09544c674c0c8df09f880)~~~


CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -51,7 +51,7 @@ function Stop-ProcessIfExists {
     Invoke-Command -Session $Session -ScriptBlock {
         $Proc = Get-Process $Using:ProcessName -ErrorAction SilentlyContinue
         if ($Proc) {
-            $Proc | Stop-Process -Force
+            $Proc | Stop-Process -Force -PassThru | Wait-Process -ErrorAction SilentlyContinue
```
kaweue, line 54: It seems to be fine

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -167,6 +167,10 @@ function Disable-DockerDriver {
         Get-ContainerNetwork | Remove-ContainerNetwork -Force
         Start-Service docker | Out-Null
     }
+
+    # Wait for Docker service to warm up in order
+    # to avoid `HNS failed with error : Unspecified error` in docker driver.
+    Start-Sleep -Seconds 6
```
kaweue: I think we shouldn't add another sleeps to test integration. We should rather handle this case correctly.

kaweue (changes requested):

CIScripts/Test/TestConfigurationUtils.ps1
kaweue: I wonder why do we have to add this SilentlyContinue ?

krdln: If `Stop-Process` kills the process fast enough, `Wait-Process` could get dead process on input.

CIScripts/Test/TestConfigurationUtils.ps1
krdln: I've investigated this a little further. So the problem is that `Docker` service creates its own VMSwitch and a container network called `nat` and while it's doing that, our attempt to create another network will result in `HNS unspecified`. I see three ways to fix that:

1. Disable Docker's `nat` network in its config on testbeds: `{
    "bridge" : "none"
}`
2. Instead of this sleep, wait for `nat` network to appear.
3. Fix this in docker driver (not sure how to do it properly yet)

m-kostrzewa: I would go for 3. 
and also write a test case in this test suite: https://github.com/codilime/contrail-windows-docker/blob/master/hns/hns_test.go#L406

krdln: I'll revert this commit then, and do the waiting in docker driver.

jablonskim (approved):

kaweue (approved):

PR #9, opened 2017-12-20
Jw 1224 tcp test fix events

krdln:
This PR doesn't fix the TCP test flakiness, but fixes error reporting, so it should produce better logs in case of another failure.

It also changes multihost udp test to use events to synchronization instead of hardcoded sleep.


kaweue (dismissed):

CIScripts/Test/Tests/VRouterAgentTests.ps1
```diff
@@ -1073,31 +1063,37 @@ function Test-VRouterAgentIntegration {
...
 
-                & docker exec $Using:Container2Name powershell -Command $Command
+                $JobListener = Start-Job -ScriptBlock {
+                    Param ($ContainerName, $Command)
+                    & docker exec $ContainerName powershell -Command $Command
+                } -ArgumentList $Using:Container2Name, $Command
             }
 
-            Start-Sleep -Seconds 8
+            Wait-RemoteEvent -Session $Session2 -ContainerName $Container2Name -EventName "Listening"
```
jacek-izykowski: Does UdpClient actually listen when ReceiveAsync() returns?

Seemingly the only thing we should rely on is that returned Task object is valid.
I suspect that we actually may still rely on delay between calling ReceiveAsync() and sending UDP packet. It's now implicit though. However, additional event is for sure a good step.

I suggest adding a comment about this uncertainty - if it gets flaky, we might add explicit delay anyway (after Wait-RemoteEvent).

jacek-izykowski (changes requested):

jacek-izykowski (approved):

PR #11, opened 2018-01-03
JW-1377 Wait for proper IP when creating agent config

krdln:
This should fix some of the flakiness in multihost tests.


CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -411,7 +411,18 @@ function New-AgentConfigFile {
         $VHostIfIndex = $Using:VHostIfIndex
         $PhysIfName = $Using:PhysIfName
 
-        $VHostIP = (Get-NetIPAddress -ifIndex $VHostIfIndex -AddressFamily IPv4).IPAddress
+        $WAIT_TIME_FOR_DHCP_IN_SECONDS = 60
```
kaweue: Could it be done by calling something similar to dhclient on windows like maybe  'ipconfig /renew' ?

For example:
You could check if there is valid address from dhcp using https://blogs.msdn.microsoft.com/virtual_pc_guy/2015/07/31/powershell-direct-waiting-for-an-ip-address/

After that you could call ipconfig /release; ipconfig /renew on selected interface.

kaweue (changes requested):

CIScripts/Test/TestConfigurationUtils.ps1
jablonskim: I think we should create new function (`Wait-InterfaceIP` maybe), move the code there (with all changes requested by @PawelMarchewkaCodiLime) and call it from `Initialize-TestConfiguration` to have valid IP address in every case.

jablonskim (changes requested):

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -442,6 +457,8 @@ function Initialize-ComputeServices {
                [Parameter(Mandatory = $false)] [Boolean] $NoNetwork = $false)
 
         Initialize-TestConfiguration -Session $Session -TestConfiguration $TestConfiguration -NoNetwork $NoNetwork
+        $HNSTransparentAdapter = Get-RemoteNetAdapterInformation -Session $Session -AdapterName $TestConfiguration.VHostName
```
kaweue: We think that this should be rather in `Initialize-TestConfiguration`, since `Initialize-ComputeServices` is not used everywhere (`Test-InitialPkt0Injection`, `Send-UDPPacket`).

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -322,6 +322,21 @@ function Remove-AllUnusedDockerNetworks {
...
+    Param ([Parameter(Mandatory = $true)] [PSSessionT] $Session,
+           [Parameter(Mandatory = $true)] [Int] $ifIndex)
+
+    Invoke-Command -Session $Session -ScriptBlock {
+        $Address = Get-NetIPAddress -InterfaceIndex $Using:ifIndex -ErrorAction SilentlyContinue `
+                    | ? AddressFamily -eq IPv4 `
+                    | ? SuffixOrigin -eq "Dhcp"
+        if (!$Address) {
+            $AdapterConfig = Get-WmiObject -Class Win32_NetworkAdapterConfiguration | ? InterfaceIndex -eq $Using:ifIndex
+            $AdapterConfig.RenewDHCPLease() | Out-Null
```
kaweue: We think we should check if it succeeded. Please look here (https://msdn.microsoft.com/en-us/library/aa393064(v=vs.85).aspx). Maybe throw an exception when return value is other than 0.

kaweue (changes requested):

krdln:
Testing revealed that `.RenewDHCPLease()` doesn't work reliably. On my computer, renewing a valid IP succeeds, but on testbeds, it fails after after about 30 seconds. Perhaps it's because the interface is used by vrouter? Anyway, I think we should go with loop here.

kaweue (approved):

m-kostrzewa:
btw please add ticket number to PR title

jablonskim (approved):

PR #13, opened 2018-01-04
Install go tools globally on builder

krdln:
This moves installation of go-msi, ginkgo and dep to ansible role
making these tools available globally on builder.

Note: Ansible script is still WIP as it contains at least one bug (robocopy src uses file path instead of directory) but we're going to merge anyway, to make builds stop failing on go-msi issue. The fix for ansible will come in next PR.


ansible/roles/builder/tasks/main.yml
```diff
@@ -170,6 +170,27 @@
     version: 1.8.3
     state: present
 
+- name: Install go-msi
+  win_chocolatey:
+    name: go-msi
+    state: present
```
wurbanski: Required version should be specified when using public chocolatey.

wurbanski (changes requested):

wurbanski (dismissed):

ansible/roles/builder/tasks/main.yml
```diff
@@ -170,6 +170,28 @@
...
 
+- name: Install go-msi
+  win_chocolatey:
+    name: go-msi
+    version: '1.0.2'
+    state: present
+
+- name: Install dep
+  win_get_url:
+    url: 'https://github.com/golang/dep/releases/download/v0.3.2/dep-windows-amd64'
```
sodar, line 181: I think that, if we download it like this, then `dep` will be available in `PATH` as `dep-windows-amd64` and not as `dep`.

sodar (changes requested):

ansible/roles/builder/tasks/main.yml
krdln: Good point

sodar (dismissed):

sodar (approved):

kaweue (approved):

krdln:
Ansible script is still WIP as it contains at least one bug (robocopy src uses file path instead of directory) but we're going to merge anyway, to make builds stop failing on go-msi issue. The fix for ansible will come in next PR.

PR #16, opened 2018-01-08
JW-1344: use zuul-cloner to checkout repositories

sodar:


ansible/roles/builder/tasks/main.yml
```diff
@@ -28,6 +39,16 @@
...
 
+- name: Install GitPython
+  win_command: pip install GitPython==2.1.8
+  args:
+    creates: C:\Python27\Lib\site-packages\git\__init__.py
+
+- name: Install Zull (for zuul-cloner)
+  win_command: pip install zuul==2.6.0
+  args:
+    creates: C:\Python27\Scripts\zuul-cloner.exe
```
m-kostrzewa: Maybe we should switch to using requirements.txt? http://docs.ansible.com/ansible/latest/pip_module.html

CIScripts/Build/Zuul.ps1
```diff
@@ -0,0 +1,39 @@
+. $PSScriptRoot\..\Common\DeferExcept.ps1
+
+function Clone-ZuulRepos {
+    $ZuulClonerOptions = @(
+        "-m",
+        "./CIScripts/clonemap.yml",
+        "--verbose",
+        $Env:GERRIT_URL
```
m-kostrzewa: Maybe it would be better to pass `GERRIT_URL` via a param? So there are no hidden globals used all over the code

ansible/roles/builder/tasks/main.yml
```diff
@@ -28,6 +39,16 @@
   args:
     creates: C:\Python27\Lib\site-packages\lxml\__init__.py
 
+- name: Install Zull (for zuul-cloner)
+  win_command: pip install zuul==2.6.0
```
krdln: Perhaps we can use the [pip module](http://docs.ansible.com/ansible/latest/pip_module.html) here? (Not sure if it works on windows though)

jacek-izykowski (changes requested):
What do you think about extracting environment variables in top-level script and passing them explicitly (e.g. GERRIT_URL)?

KR, JI

jacek-izykowski (approved):
KR, JI

kaweue (approved):

PR #20, opened 2018-01-09
JW-1378 DeferExcept: Detect error when command doesn't run at all

krdln:
Now it detects eg. command not found error.

Note: After this change, empty block will not be considered successful.


krdln:
It seems that this approach doesn't work.

kaweue (approved):

jablonskim (approved):

PR #22, opened 2018-01-11
Move create tester template and deploy tester to jenkinsfiles

katrybacka:


ansible/roles/tester/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
+---
+
+- name: Prepare Jenkins partition
+  win_shell: 'Get-Disk |
+  Where partitionstyle -eq "raw" |
+  Initialize-Disk -PartitionStyle MBR -PassThru |
+  New-Partition -DriveLetter J -UseMaximumSize |
+  Format-Volume -FileSystem NTFS -NewFileSystemLabel "Jenkins" -Confirm:$false'
```
krdln: Maybe use "folded string style" here? See https://symfony.com/doc/current/components/yaml/yaml_format.html#strings

vars/prepareVMConfig.groovy
```diff
@@ -0,0 +1,38 @@
...
+  if(vm_hdd_size != null){
+    vmConfig += """
+vm_hdd:
+  - type: thin
+    size: ${vm_hdd_size}
+"""
+  }
+
+  writeFile file: file_path, text: vmConfig
+}
```
krdln: Note: this is just a suggestion

I'd remove this file. Instead, I'd create a dictionary in groovy, and pass it as `extraVars` parameter of `ansiblePlaybook` step.

See this example https://github.com/jenkinsci/ansible-plugin#example
(We'd need to create this dictionary in particular Jenkinsfile's and pass them as parameters here)

katrybacka: As far as I know extraVars won't work with nested lists or maps

vars/prepareVMConfig.groovy
```diff
@@ -0,0 +1,38 @@
...
+  memory_mb: ${vm_hardware_memory_mb}
+  num_cpus: ${vm_hardware_num_cpus}
+
+vm_networks:
+"""
+
+  for(network in vm_networks){
+    network.eachWithIndex{key, value, index -> if(index == 0) vmConfig += "  - "
+                                               else vmConfig += "    "
+                                               vmConfig += "$key: $value\n"}
```
krdln: You can use alternative syntax for maps in yaml, to avoid this `if(index == 0)` hackery. ~~~Ctrl+F `json_map` on https://learnxinyminutes.com/docs/yaml/~~~
http://docs.ansible.com/ansible/latest/YAMLSyntax.html

I mean, instead of using
```
   - key1: foo
     key2: bar
```
You can use
```
   - { key1: "foo", key2: "bar" }
```

krdln (changes requested):

vars/prepareVMConfig.groovy
krdln: Also, please use spaces around `()` in for.

katrybacka: in if statement also

ansible/roles/tester/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
+---
+
+- name: Prepare Jenkins partition
+  win_shell: 'Get-Disk |
```
sodar: We can extract this script to another role, e.g. `Jenkins-Partition` and include it here and in builder playbooks

vars/prepareAndRunAnsiblePlaybook.groovy
```diff
@@ -17,8 +21,9 @@ def call(Map params) {
             sh 'cp inventory.sample inventory'
             sh 'ansible-galaxy install -r requirements.yml -f'
             script {
-              prepareConfig(env.VC_HOSTNAME, env.VC_DATACENTER, env.VC_CLUSTER, env.VC_FOLDER,
-                            env.VC_NETWORK, env.VC_USR, env.VC_PSW, env.VM_TEMPLATE)
+              prepareVMConfig(env.VC_HOSTNAME, env.VC_DATACENTER, env.VC_CLUSTER, env.VC_FOLDER,
```
sodar: Maybe we pass parameters through a map or keyword arguments to `prepareVMConfig`? We can reuse `params` map, but extended with other necessary parameters.

sodar (changes requested):
We came up with an idea for a little refactoring. We can divide `prepareVMConfig` into 2 functions: `getVMwareConfig` which returns a map with credentials and `prepareHardwareConfig` which creates a file with `vm_networks`, `vm_hdd`, `vm_hardware`. Map returned by `getVMwareConfig` will be passed through `extraVars` to `ansiblePlaybook` step.

jenkinsfiles/Jenkinsfile.deploy-builder
```diff
@@ -1,4 +1,7 @@
 library "contrailWindows@$BRANCH_NAME"
 
-prepareAndRunAnsiblePlaybook(config: prepareDeployBuilder,
-                             playbook: 'vmware-deploy-template.yml')
+prepareAndRunAnsiblePlaybook(playbook: 'vmware-deploy-template.yml',
+                             vm_role: 'builder',
```
m-kostrzewa: Do we really need to specify all these params to deploy? Arent they stored in template already?

ansible/roles/jenkins-partition/tasks/main.yml
```diff
@@ -0,0 +1,10 @@
+---
```
krdln, line 1: Yaml files should begin with `---`, but also end with `...` (see http://docs.ansible.com/ansible/latest/YAMLSyntax.html)

It seems those markers are not necessary, but if we're adding a starting marker, let's add also the ending marker. (also applies to other tasks).

m-kostrzewa (dismissed):

vars/getBaseHardwareConfig.groovy
```diff
@@ -0,0 +1,16 @@
+def call(vm_networks, vm_hardware_memory_mb, vm_hardware_num_cpus) {
+
+  def nets = vm_networks.inject([]) { result, network ->
+                                      result << "  - {${ network.collect().join(', ') }}"
+                                    }.join('\n').replaceAll("=", ": ")
```
krdln: Nice! Although it feels a little bit too magic. I'd propose:
1. Create a helper function (in this file) eg. `mapToYaml` that converts a `Map` to yaml syntax (basically `network.collect().join(', ')`
2. Change inject to just `collect` and simply `join('')` (so move `\n` from `join` to the closure)
3. Move `replaceAll` to `mapToYaml`, so it's more obvious what generates the `=` which we're converting

vars/getVMwareConfig.groovy
```diff
@@ -0,0 +1,14 @@
+def call(vm_role) {
+
+  def baseVMParams = [vcenter_hostname: env.VC_HOSTNAME,
+                      vcenter_user: env.VC_USR,
+                      vcenter_password: env.VC_PSW,
+                      validate_certs: 'no',
+                      datacenter_name: env.VC_DATACENTER,
+                      cluster_name: env.VC_CLUSTER,
+                      vmware_folder: (env.VC_FOLDER).replaceAll("\"", ""),
```
krdln: Unnecessary `()` around `env.VC_FOLDER` (not sure though)

vars/prepareAndRunAnsiblePlaybook.groovy
```diff
@@ -1,7 +1,8 @@
 def call(Map params) {
+  def prepareHardwareConfig = params.config
   def playbook = params.playbook
-  def prepareConfig = params.config
-
+  def vm_role = params.vm_role
+  def vmWareConfig
```
krdln, line 5: We can call `getVMwareConfig` directly in `ansiblePlaybook` step, and remove this global variable.

krdln (changes requested):

vars/prepareAndRunAnsiblePlaybook.groovy
katrybacka: I placed this function in the stage of preparing environment for running ansible, because i think that whole environment should be prepared for running ansible in this stage and stage "run ansible" should be only responsible for executing playbook.

vars/getBaseHardwareConfig.groovy
katrybacka: We have here list of maps, each map represents network. As far as I know using ```join``` change whole collection into String so basically you cannot use ```join``` twice on nested collection.

ansible/roles/jenkins-partition/tasks/main.yml
katrybacka: I'm not sure if it is a part of this task to change our convention. I can change it here, but I'm not sure if it is necessary and futhermore changes in other files should be done in another PR.

vars/getBaseHardwareConfig.groovy
krdln: Now you have one `join`inside the closure and one outside. I didn't mean to change that

ansible/roles/jenkins-partition/tasks/main.yml
krdln: I've noticed it only on new files, but you're right, we can change the rest in a separate PR

vars/prepareAndRunAnsiblePlaybook.groovy
krdln: Seems reasonable then

vars/getBaseHardwareConfig.groovy
katrybacka: But by using ```inject``` method I still have a list, just containing desired strings generated from maps. Without that I would have one string and second operation ```join``` would not succeed.

vars/createTemplateConfig.groovy
```diff
@@ -0,0 +1,17 @@
+def call(params, file_path = 'vm.vars') {
+
+  def vm_networks = params.vm_networks
+  def vm_hardware_memory_mb = params.vm_hardware_memory_mb
+  def vm_hardware_num_cpus = params.vm_hardware_num_cpus
+  def vm_hdd_size = params.vm_hdd_size
```
sodar, line 6: Can you check if when `vm_hdd_size` does not exist in `params` map, it will throw an exception?

sodar (approved):
LGTM, one minor comment.

DS, MKr

krdln (approved):

PR #26, opened 2018-01-15
JW-1290 Copy deploy keys to builders and use them to clone Docker Driver

krdln:
I haven't changed the cloning of windowsstubs yet (the key and ssh config is ready though)


krdln:
retest this please

ansible/roles/builder/defaults/main.yml
```diff
@@ -1,2 +1,3 @@
 dependencies_dir: 'C:\BUILD_DEPENDENCIES'
 dependencies_source: '\\{{ hostvars[groups["drive"][0]]["inventory_hostname"]}}\SharedFiles'
+git_home_dir: 'C:\Windows\system32\config\systemprofile'
```
kaweue: We think we could use different more generic name than git_home_dir (like system_profile_dir).

Is it possible to get this path from some Windows env variable ?

kaweue (changes requested):
PM + KR

katrybacka (changes requested):
KR

ansible/roles/builder/defaults/main.yml
krdln: This is a funny story: The homedir is usually determined from `$HOMEDRIVE` and `$HOMEPATH` env variables (so eg. `C:\Users\Administartor`), but these variables are *not* set when Jenkins stage is running on builder.

I determined this path empirically, by checking where git looks for `.git` and `.ssh` directory. I've named this variable as `git_home_dir`, because I'm not sure whether other applications use this directory too, or is it a git specific fallback.

kaweue: We accept to use path directly as it is. However C:\Windows\system32\config\systemprofile is also used by other applications, we have checked that on my windows. We suggest to change it to more generic one.

krdln: I'm not sure about naming `system_profile_dir`, because it describes contents of the variable rather that intent. For example, if for some reason Jenkins started to set `$HOMEPATH` to eg. `C:\Users\Jenkins`, git would look for its config there and the name `system_profile_dir` would not fit, because it would be a user profile dir rather than system one. Perhaps `user_profile_dir` then?

I'll commit renaming to `system_profile_dir` in case we don't find a good name :)

kaweue (approved):

katrybacka (approved):

PR #29, opened 2018-01-16
JW-1384 - Adding support for production and development pipelines

kaweue:


ansible/roles/zuul/files/zuul/layout.yaml
```diff
@@ -15,35 +14,54 @@ pipelines:
...
         - event: comment-added
           comment: (?i)recheck windows$
+  - name: development
+    description: Pipeline used for Windows CI development. It should always work in silent mode.
+    manager: IndependentPipelineManager
+    precedence: low
+    trigger:
+      gerrit:
+        - event: comment-added
+          comment: (?i)recheck windows-exp$
```
kaweue: I think we should use: "recheck windows-dev"

krdln: Perhaps `(?:-| )` so one doesn't have to remember whether it was `-` or space.
I'd also add a `\s*` before `$` in each comment, for the trailing whitespace to be ignored.

ansible/roles/zuul/files/zuul/layout.yaml
```diff
@@ -15,35 +14,54 @@ pipelines:
...
+    development:
+      - '{jobprefix}-development'
 
 projects:
   - name: Juniper/contrail-controller
-    check:
-      - ci-contrail-windows
+    template:
+      - name: ci-contrail-windows-template
+        jobprefix: ci-contrail-windows
```
krdln: If `jobprefix` is always the same, I don't see a point in passing it as a parameter to template.

krdln (changes requested):

ansible/roles/zuul/files/zuul/layout.yaml
```diff
@@ -15,35 +14,54 @@ pipelines:
...
         - event: comment-added
           comment: (?i)recheck windows$
+  - name: development
+    description: Pipeline used for Windows CI development. It should always work in silent mode.
+    manager: IndependentPipelineManager
+    precedence: low
+    trigger:
+      gerrit:
+        - event: comment-added
+          comment: (?i)recheck windows(?:-| )exp\s*$
```
kaweue: should be "recheck windows-dev" - not fixed yet

krdln: Since you haven't changed that yet, maybe `dev|devel|development`?

ansible/roles/zuul/files/zuul/layout.yaml
jacek-izykowski: This is actually template (pattern) itself that is changing (production/development). Do we need a template to achieve this effect (maybe we do - I'm just asking)?

jacek-izykowski (changes requested):

krdln (approved):

jacek-izykowski (approved):

PR #34, opened 2018-01-17
JW-1382 Fix failed stages appearing as successful and rename DeferExcept

krdln:
The wrapper used by Jenkins to run powershell scripts look at `$LastExitCode`, so if last command had set it to 0, and then powershell command would fail, then stage would be marked as successful. To fix that, we set `$LastExitCode` to `$null`.

This PR also renames `DeferExcept` to `Invoke-NativeCommand` to better describe purpose of this function.

**Note**: Every native command should be run using this function, or it will confuse Jenkins.

**Note**: Only places allowed to use `$LastExitCode` now are:
* `Invoke-NativeCommand` itself
* Inside of the `ScriptBlock` passed to `Invoke-NativeCommand`
* On remote machines (inside of `Invoke-Command -Session $Session`)


krdln:
[tested in juniper CI](http://10.84.12.26:8080/job/WinContrail/job/winci-mkr-test/39/console)

krdln:
The build-all job [failed](http://ci.wincontrail.codilime.com:8080/job/Wincontrail/job/build-all/1596/console), but it seems that all the changes from this PR are working correctly.

sodar (approved):

jacek-izykowski (approved):

PR #35, opened 2018-01-17
Global agent should be set to none and used explicitly in Post stage.

kaweue:


kaweue:
http://10.84.12.26:8080/job/WinContrail/job/WinCI-pm-test/6/console - It is working

krdln (approved):

PR #31, opened 2018-01-16
TestConfiguration from file

m-kostrzewa:

PR #42, opened 2018-01-19
JW-1398 Setup TestConfiguration in file instead of envs 

krdln:


Jenkinsfile
```diff
@@ -75,6 +75,8 @@ pipeline {
             environment {
                 // TESTBED_HOSTNAMES = SpawnedTestbedVMNames
                 ARTIFACTS_DIR = "output"
+                // TODO actually create this file
+                TEST_CONFIGURATION_FILE = "GetTestConfigurationJuni.ps1"
```
m-kostrzewa, line 79: Since we're abandonning Codi CI I think we can keep it in code.

CIScripts/Test/TestRunner.ps1
```diff
@@ -17,71 +17,18 @@
 . $PSScriptRoot\Tests\WindowsLinuxIntegrationTests.ps1
 
 function Run-TestScenarios {
-    Param ([Parameter(Mandatory = $true)] [PSSessionT[]] $Sessions)
+    Param (
+        [Parameter(Mandatory = $true)] [PSSessionT[]] $Sessions,
+        [Parameter(Mandatory = $true)] [String] $TestConfigurationFile
```
m-kostrzewa, line 22: I don't see the reason to pass `$TestConfigurationFile` instead of `$TestConfiguration` struct here. Loading config is one responsibility of `Run-TestScenarios` that we can get rid of here thanks to our refactor (which is good - SRP). What do you think?

CIScripts/Test/TestRunner.ps1
```diff
@@ -17,71 +17,18 @@
...
-            Name = $Env:NETWORK_WITH_POLICY_1_NAME
-            Subnets = @($Env:NETWORK_WITH_POLICY_1_SUBNET)
-        }
+        . "$TestConfigurationFile"
 
-        $NetworkWithPolicy2Configuration = [NetworkConfiguration] @{
-            Name = $Env:NETWORK_WITH_POLICY_2_NAME
-            Subnets = @($Env:NETWORK_WITH_POLICY_2_SUBNET)
-        }
+        $TestConfiguration = Get-TestConfiguration
```
m-kostrzewa, line 29: So, we use a consolidation seam here (we source a file with specific implementation of functions). I'm not sure I like it tbh, it generates additional 

Why not just use the method in this PR: https://github.com/codilime/contrail-windows-ci/pull/31/files ? Then we have something like `$TestConfiguration = Import-LocalizedData -FileName TestConfiguration.psd1`.

m-kostrzewa (dismissed):

CIScripts/Test/TestRunner.ps1
krdln: The only problem is that the file actually constains 3 test configurations, so we could pass either 3 `TestConfigurations` or create something like `TestConfigurationProvider`. What do you think about that?

CIScripts/Test/TestRunner.ps1
krdln: >  it generates additional

Additional what?

I've tried to make it work before with `Import-LocalizedData` and concluded that it's really not an appropriate way to do it, because:
1. It insists on presence of subdirectories like `en_US`.
2. The name of the function is confusing
3. This file can contain map of maps, we need nested typed objects. (Perhaps `psd1` files can contain objects. I'll need to retest that)
4. We want to load 3 similar, but different `TestConfiguration`s. The easiest way to describe "take the first test configuration and change some variables" is with code.

In general, my idea was to create an interface with 3 functions (`Get-TestConfiguration`, -`Udp` and -`WindowsLinux`) and select the provider with environment variables. You can imagine that we're just selecting implementation of this interface with a file.

Jenkinsfile
krdln: No sure what you mean here.

CIScripts/Test/TestRunner.ps1
m-kostrzewa: There's no need for `GetTestConfigurationCodiLegacy.ps1` and `GetTestConfigurationJuni.ps1`. One file would be enough.

Then we don't need to source file inside `Run-TestScenarios` func - we can include it on the top of the file like normal.

m-kostrzewa:
TBH this changeset only makes me realize how bad the current usage of `$TestConfiguration` struct is. 
Maybe we should focus on refactoring the usage of it first.

So maybe we should just put every env var in Jenkinsfile for now, so we don't complicate the refactor. And then refactor...

krdln:
The current plan is to get rid of the Env variable and just move these function to some regular file. @jacek-izykowski

sodar (approved):

krdln:
recheck no bug

PR #47, opened 2018-01-23
JW-1387 Export logs from Jenkins to log server

jacek-izykowski:


Jenkinsfile
```diff
@@ -90,8 +90,11 @@ pipeline {
                 // cleanWs()
                 sh 'echo "TODO environment cleanup"'
                 // unstash "buildLogs"
-                // TODO correct flags for rsync
-                sh "echo rsync logs/ logs.opencontrail.org:${JOB_NAME}/${BUILD_ID}"
+                sh """#!/bin/bash
```
jablonskim: 1. Please move `mkdir` to separate `sh` command. Use `ssh` to call it on remote host.
2. Move all `rsync` parameters to seperate variables (maybe use `environment` section).
3. Try to use `replaceAll` method from groovy for substitution.
4. Remove `#!/bin/bash`.
5. For refactoring purposes, please change all occurences of `sh 'echo "..."'` to `echo '...'`.

KR & MJ

jablonskim (changes requested):

kaweue:
recheck no bug

Jenkinsfile
jacek-izykowski: 1. Done (--rsync-path is actually meant as a tool to initialize target directory as it was done though).
2. Done (new variables are only relevant within shell invocation, so they're defined there).
3. Now groovy substitution is not used at all.
4. I can't - it's actually needed (for variable substitution).
5. Done (for 'post' section).

katrybacka (approved):

jablonskim (approved):

PR #51, opened 2018-01-25
RFC Pester

m-kostrzewa:
Proposed refactor (POC) for Pester.

Benefits:
* Can generate NUnit test output
* Pretty nice output
* IMHO nicer syntax
* Can run tests separately (test script only requires testbed address, creds and config)

This PR is based on @krdln's work to move configuration to a file.

Main changes:
* Removed `Pkt0PipeImplementationTests.ps1` - those tests are now in Pester syntax in `Pkt0.Tests.ps1`
* Added `Invoke-Pester` to test runner

TODO: all other tests :)


CIScripts/Test/TestRunner.ps1
```diff
@@ -45,7 +45,9 @@ function Run-TestScenarios {
         # Test-DockerDriverMultiTenancy -Session $Sessions[0] -TestConfiguration $TestConfiguration
         Test-WindowsLinuxIntegration -Session $Sessions[0] -TestConfiguration (Get-TestConfigurationWindowsLinux)
         # Test-Pkt0PipeImplementation -Session $Sessions[0] -TestConfiguration $TestConfiguration
-        Invoke-Pester -Script @{Path=".\Test\Tests\Pkt0.Tests.ps1"; Parameters=@{TestbedAddr=$Sessions[0].ComputerName}}
+        if($Env:RUN_PESTER_TESTS -eq "1") {
+            Invoke-Pester -Script @{Path=".\Test\Tests\Pkt0.Tests.ps1"; Parameters=@{TestbedAddr=$Sessions[0].ComputerName}}
```
krdln: When some tests fail, how this failure is propagated up?

Also, make sure if `Invoke-Pester` doesn't set `$LastExitCode`. If it does, it has to be wrapped in `Invoke-NativeCommand` (why? tldr: Jenkins pipeline powershell wrapper is weird)

CIScripts/Test/Tests/Pkt0.Tests.ps1
```diff
@@ -0,0 +1,81 @@
...
+. $PSScriptRoot\..\..\Common\VMUtils.ps1
+
+. $PSScriptRoot\..\GetTestConfigurationCodiLegacy.ps1
+$TestConf = Get-TestConfiguration
+$Session = New-PSSession -ComputerName $TestbedAddr -Credential (Get-VMCreds)
+
+Describe "vRouter Agent" {
+    Context "given vRouter Forwarding Extension is NOT running" {
+        It "crashes" {
+            Assert-AgentProcessCrashed -Session $Session -After $TimeBeforeCrash | Should Throw
```
krdln: 1. Actually, it should *not* throw
2. Left hand side of `|` should be wrapped in `{}`
3. As the syntax `{ Assert ... } | Should Not Throw` is easy to get wrong (as this example shows), I'd go with a plain `Assert` _for now_

CIScripts/Test/Tests/Pkt0.Tests.ps1
```diff
@@ -0,0 +1,81 @@
...
+                | Should Be $false
+            $TimeBeforeCrash = Invoke-Command -Session $Session -ScriptBlock { Get-Date }
+            Enable-AgentService -Session $Session
+        }
+    }
+
+    Context "given vRouter Forwarding Extension is running" {
+        It "runs correctly" {
+            $Result = Assert-IsAgentServiceEnabled -Session $Session
+            $Result | Should BeNullOrEmpty
```
krdln: `Should BeNullOrEmpty` is unnecessary -- `Assert-IsAgentServiceEnabled` either returns null or throws. (the return inside of it is inside a ScriptBlock and can be removed.

CIScripts/Test/Tests/Pkt0.Tests.ps1
```diff
@@ -0,0 +1,81 @@
...
+                -VMSwitchName $TestConf.VMSwitchName `
+                -ForwardingExtensionName $TestConf.ForwardingExtensionName `
+                | Should Be $true
+            Enable-AgentService -Session $Session
+        }
+    }
+
+    Context "vRouter Forwarding Extension was disabled while Agent was running" {
+        It "crashes" {
+            Assert-AgentProcessCrashed -Session $Session -After $TimeBeforeCrash | Should Throw
```
krdln: Same as in first test.

CIScripts/Test/Tests/Pkt0.Tests.ps1
```diff
@@ -0,0 +1,81 @@
...
+)
+
+. $PSScriptRoot\..\Utils\CommonTestCode.ps1
+. $PSScriptRoot\..\TestConfigurationUtils.ps1
+. $PSScriptRoot\..\..\Common\Aliases.ps1
+. $PSScriptRoot\..\..\Common\VMUtils.ps1
+
+. $PSScriptRoot\..\GetTestConfigurationCodiLegacy.ps1
+$TestConf = Get-TestConfiguration
+$Session = New-PSSession -ComputerName $TestbedAddr -Credential (Get-VMCreds)
```
krdln: Perhaps add option (I'd even say that it should be default) to run without `-Credential`, so it's even easier to test locally.

krdln (changes requested):

CIScripts/Test/Tests/Pkt0PipeImplementationTests.ps1
```diff
@@ -1,114 +0,0 @@
-function Test-Pkt0PipeImplementation {
```
krdln, line 1: What happened to this file? (I don't see a Pester version of it too)

CIScripts/Test/TestRunner.ps1
```diff
@@ -30,27 +29,36 @@ function Run-TestScenarios {
...
+        # Test-MultipleSubnetsSupport -Session $Sessions[0] -TestConfiguration $TestConfiguration
+        # Test-DockerDriverMultiTenancy -Session $Sessions[0] -TestConfiguration $TestConfiguration
+        #Test-WindowsLinuxIntegration -Session $Sessions[0] -TestConfiguration (Get-TestConfigurationWindowsLinux)
+        # Test-Pkt0PipeImplementation -Session $Sessions[0] -TestConfiguration $TestConfiguration
+        
+        # if($Env:RUN_DRIVER_TESTS -eq "1") {
+        #     Test-DockerDriver -Session $Sessions[0] -TestConfiguration $TestConfiguration
+        # }
     })
+    if($Env:RUN_PESTER_TESTS -eq "1") {
```
krdln: Please future-proof the code for strict mode by adding `(Test-Path Env:RUN_PESTER_TESTS) -and `

Please add space after `if` (here and in all the other places)

CIScripts/Test/TestRunner.ps1
```diff
@@ -30,27 +29,36 @@ function Run-TestScenarios {
...
+        $TestPath = (Get-ChildItem -Recurse -Filter "vRouterAgent.Tests.ps1")[0].FullName
+        $Results = Invoke-Pester -PassThru -Script @{Path=$TestPath; `
+                Parameters=@{TestbedAddr=$Sessions[0].ComputerName}; Arguments=@()}
+        echo "Num failed tests: $($Results.FailedCount)"
```
krdln: Pleas use `Write-Host` instead of `echo`.

CIScripts/Test/Tests/vRouterAgent.Tests.ps1
```diff
@@ -0,0 +1,131 @@
+Param (
```
krdln: Why this file is named `vRouterAgent.Tests`? Shouldn't it be `AgentService.Tests`? In current test we have `vRouterAgentTest.ps1` which tests vRouter-Agent integration, so this name will be confusing.

CIScripts/Test/Tests/vRouterAgent.Tests.ps1
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    Context "uninstalling using MSI" {
+        It "unregisters Agent service" {
+            Assert-IsAgentServiceUnregistered -Session $Session
+        }
+
+        BeforeEach {
+            Assert-IsAgentServiceRegistered -Session $Session
+            Uninstall-Agent -Session $Session
```
krdln: How this interacts with `Uninstall-Agent` that's in global `AfterEach`? (I guess second uninstallation just fails silently (as we don't check exitcode of msiexec). Do we want to keep it that way?

krdln (changes requested):

CIScripts/Test/Tests/vRouterAgent.Tests.ps1
m-kostrzewa: `vRouterAgentTest.ps1` will probably be renamed in the future.

 > Shouldn't it be `AgentService.Tests`?

Not sure, since we test MSI as well.

CIScripts/Test/Tests/Pkt0PipeImplementationTests.ps1
m-kostrzewa: Removed it

CIScripts/Test/Tests/vRouterAgent.Tests.ps1
krdln: I'd name it just `Agent.Tests`, so it won't get confused with current `vRouterAgentTest.ps1`.

CIScripts/Test/TestRunner.ps1
```diff
@@ -1,6 +1,6 @@
 . $PSScriptRoot\TestConfigurationUtils.ps1
+. $PSScriptRoot\..\Common\Invoke-NativeCommand.ps1
```
krdln: No longer needed.

CIScripts/Test/TestRunner.ps1
```diff
@@ -30,27 +29,40 @@ function Run-TestScenarios {
...
+        #Test-WindowsLinuxIntegration -Session $Sessions[0] -TestConfiguration (Get-TestConfigurationWindowsLinux)
+        # Test-Pkt0PipeImplementation -Session $Sessions[0] -TestConfiguration $TestConfiguration
+        
+        # if($Env:RUN_DRIVER_TESTS -eq "1") {
+        #     Test-DockerDriver -Session $Sessions[0] -TestConfiguration $TestConfiguration
+        # }
     })
+    if ($Env:RUN_PESTER_TESTS -eq $true) {
+        $TotalFailedCount = 0
+        Get-ChildItem -Recurse -Filter "*.Tests.ps1" | ForEach-Object {
```
krdln: Use a regular `foreach` loop here.
I had a conversation with @PawelMarchewkaCodiLime and the conclusion was that `ForEach-Object` aka `%` is error-prone and we should avoid it in our codebase (unless the collection is really large and memory usage is too high).

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -245,6 +269,32 @@ function Assert-IsAgentServiceDisabled {
     throw "Agent service is not disabled. EXPECTED: Agent service is disabled"
 }
 
+function Test-IsAgentServiceRegistered {
+    Param ([Parameter(Mandatory = $true)] [System.Management.Automation.Runspaces.PSSession] $Session)
+
+    $Service = Invoke-Command -Session $Session -ScriptBlock {
+        return $(if (Get-Service "ContrailAgent" -ErrorAction SilentlyContinue) { $true } else { $false })
+    }
+    return $Service
```
krdln: These copy-pasted functions could get some more love. For example `return $()` is totally unnecessary here. Moreover, the `$Service` variable is also not necessary here. The whole function could be reduced to a single `Invoke-Command`.

(I know you just wanted to copy-paste, but I guess a little bit of clean up won't hurt :))

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -245,6 +269,32 @@ function Assert-IsAgentServiceDisabled {
...
+}
+
+function Assert-IsAgentServiceRegistered {
+    Param ([Parameter(Mandatory = $true)] [System.Management.Automation.Runspaces.PSSession] $Session)
+
+    $Res = Test-IsAgentServiceRegistered -Session $Session
+    if (!$Res) {
```
krdln: This `$Res` variable could have more meaningful name (or maybe just get rid of it)

CIScripts/Test/Tests/vRouterAgentMSIInstaller.Tests.ps1
```diff
@@ -0,0 +1,32 @@
...
+. $PSScriptRoot\..\..\Common\VMUtils.ps1
+
+. $PSScriptRoot\..\GetTestConfigurationCodiLegacy.ps1
+$TestConf = Get-TestConfiguration
+$Session = New-PSSession -ComputerName $TestbedAddr -Credential (Get-VMCreds)
+
+Describe "vRouter Agent MSI installer" {
+    Context "installation and uninstallation" {
+        It "registers/unregisters agent service and never enables" {
+            Install-Agent -Session $Session
```
krdln, line 18: The Agent is installed by default, so we're not actually testing installation here. Maybe just copy-paste the body of this test twice?

krdln (changes requested):

CIScripts/Test/Tests/vRouterAgentMSIInstaller.Tests.ps1
krdln: Or maybe reorder: first uninstall, then install

m-kostrzewa: We discussed it with @jablonskim - we won't install artifacts by default in "Provision" stage.

CIScripts/Test/TestRunner.ps1
m-kostrzewa: Interesting. I'll do it but could you link an explanation?

CIScripts/Test/TestConfigurationUtils.ps1
m-kostrzewa: Hmm, if I was going to refactor those functions, I would go all-out with the Pester's `Should`s etc. And we discussed that we should use asserts for now, until we migrate more tests.

krdln: Agree on `Should`. But if we're leaving the `Test` version anyway, I'd still clean it up from unnecessary syntax noise:
```
        if (Get-Service "ContrailAgent" -ErrorAction SilentlyContinue) { $true } else { $false }
}
```

m-kostrzewa: I will clean it up when I know how to make it work using Pester, e.g. I will probably rewrite the above line as something like

`Get-Service "ContrailAgent" | Should Not BeNull`

That's why I don't want to touch and clean up this code yet.

krdln (dismissed):
MK (just me, not pair-reviewed yet)

CIScripts/Test/TestRunner.ps1
```diff
@@ -30,27 +28,41 @@ function Run-TestScenarios {
...
+        # # TODO: Uncomment after JW-1129
+        # # Test-SNAT -Session $Sessions[0] -SNATConfiguration $SNATConfiguration -TestConfiguration $TestConfiguration
+        # Test-VRouterAgentIntegration -Session1 $Sessions[0] -Session2 $Sessions[1] `
+        #     -TestConfiguration $TestConfiguration -TestConfigurationUdp (Get-TestConfigurationUdp)
+        # Test-ComputeControllerIntegration -Session $Sessions[0] -TestConfiguration $TestConfiguration
+        # Test-MultipleSubnetsSupport -Session $Sessions[0] -TestConfiguration $TestConfiguration
+        # Test-DockerDriverMultiTenancy -Session $Sessions[0] -TestConfiguration $TestConfiguration
+        #Test-WindowsLinuxIntegration -Session $Sessions[0] -TestConfiguration (Get-TestConfigurationWindowsLinux)
+        # Test-Pkt0PipeImplementation -Session $Sessions[0] -TestConfiguration $TestConfiguration
+
```
kaweue: Extra white spaces ?

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -245,6 +269,30 @@ function Assert-IsAgentServiceDisabled {
     throw "Agent service is not disabled. EXPECTED: Agent service is disabled"
 }
 
+function Test-IsAgentServiceRegistered {
+    Param ([Parameter(Mandatory = $true)] [System.Management.Automation.Runspaces.PSSession] $Session)
+
+    $Service = Invoke-Command -Session $Session -ScriptBlock {
```
kaweue: Service variable may be misleading. I would suggest something like $IsServiceRegistered

CIScripts/Test/Tests/vRouterAgentMSIInstaller.Tests.ps1
```diff
@@ -0,0 +1,32 @@
...
+            Assert-IsAgentServiceRegistered -Session $Session
+            Assert-IsAgentServiceDisabled -Session $Session
+            Uninstall-Agent -Session $Session
+            Assert-IsAgentServiceUnregistered -Session $Session
```
kaweue: In my personal opinion opinion we should rather use pester's " | should be something", since we are trying to verify whether pester is correct solution for us.

kaweue (changes requested):

m-kostrzewa:
@krdln @PawelMarchewkaCodiLime Actually, I think I want to refactor this further to implement something similar to gomega's `Eventually` and `Consistently` (https://onsi.github.io/gomega/#making-asynchronous-assertions) so that we can write stuff like:

```
(enabling agent should make it running test)
Get-AgentServiceStatus -Session $Session | Eventually | Should Be "Running"
```

```
(agent does not restart test)
Get-AgentServiceStatus -Session $Session | Consistently | Should Not Be "Running"
```

thoughts?

edit: but in another PR, let's get this merged first

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -227,93 +229,52 @@ function Uninstall-Agent {
...
-        $Status = Invoke-Command -Session $Session -ScriptBlock {
-            return $((Get-Service "ContrailAgent" -ErrorAction SilentlyContinue).Status)
-        }
-        if ($Status.Value -eq "Running") {
-            return
-        }
-
-        Start-Sleep -s $TimeBetweenChecksInSeconds
+    $Status = Invoke-Command -Session $Session -ScriptBlock {
+        return $((Get-Service "ContrailAgent").Status)
```
krdln: Unnecessary `$()`. I think this `$()` should only be used in string interpolation. Outside a string, it's a no-op.

(The `return` and temporary `$Status` are also unnecessary). Remember, PS is quite an expression oriented language.

CIScripts/Common/Retry-Command.ps1
```diff
@@ -0,0 +1,32 @@
...
+    #
+    # Example usage:
+    # { Get-AgentServiceStatus -Session $Session } `
+    #     | Retry-Command -IntervalInSecs 1 -TimeoutInSecs 5 `
+    #     | Should Be "Stopped"
+    #
+
+    $StartTime = Get-Date
+
+    while(((Get-Date) - $StartTime).Seconds -lt $TimeoutInSecs) {
```
krdln: Please add space after `while`/`if` etc.

CIScripts/Common/Retry-Command.ps1
```diff
@@ -0,0 +1,32 @@
...
+        try {
+            $ReturnVal = & $ScriptBlock
+            return $ReturnVal
+        } catch {
+            Write-Host -NoNewline "."
+            Start-Sleep -s $IntervalInSecs
+        }
+    }
+    Write-Host ""
+    throw "Retry-Command: command timed out."
```
krdln: If we're not gonna go `Eventually` way, we should at least include the `ScriptBlock` in the thrown message, so we know what command has failed.

CIScripts/Common/Retry-Command.ps1
```diff
@@ -0,0 +1,32 @@
...
+    # { Get-AgentServiceStatus -Session $Session } `
+    #     | Retry-Command -IntervalInSecs 1 -TimeoutInSecs 5 `
+    #     | Should Be "Stopped"
+    #
+
+    $StartTime = Get-Date
+
+    while(((Get-Date) - $StartTime).Seconds -lt $TimeoutInSecs) {
+        try {
+            $ReturnVal = & $ScriptBlock
```
krdln: Missing `Write-Host ""` before return (so the next printed line is not prepended by dots)

CIScripts/Test/Tests/vRouterAgentMSIInstaller.Tests.ps1
```diff
@@ -15,10 +15,12 @@ Describe "vRouter Agent MSI installer" {
...
             Install-Agent -Session $Session
-            Assert-IsAgentServiceRegistered -Session $Session
-            Assert-IsAgentServiceDisabled -Session $Session
+            Get-AgentServiceStatus -Session $Session `
+                | Should Be "Stopped"
+                
             Uninstall-Agent -Session $Session
-            Assert-IsAgentServiceUnregistered -Session $Session
+            {Get-AgentServiceStatus -Session $Session} `
+                | Should Throw
```
krdln: I think it would be more elegant to do something like:

```powershell
Get-AgentService | Should Be NullOrEmpty
```

CIScripts/Test/Tests/vRouterAgentService.Tests.ps1
```diff
@@ -24,12 +26,14 @@ Describe "vRouter Agent service" {
...
+            Get-AgentServiceStatus -Session $Session `
+                | Should Be "Stopped"
         }
 
         It "does not restart" {
-            Start-Sleep -s 2
-            Assert-IsAgentServiceDisabled -Session $Session
+            {{ "Running" -eq (Get-AgentServiceStatus -Session $Session) } `
+                | Retry-UntilTrue -IntervalInSecs 0.2 -TimeoutInSecs 2 } `
+                | Should Throw
```
krdln: IMHO these nested piped blocks are hard to understand. Perhaps consider passing them explicitely by `-ScriptBlock` or change indentation to be more obvious? It feels like we try shoe-horn this test into functions that we have (so definitely, `Consistently` would be nicer, even writing our own `Consistently` would be clearer too)

> `-IntervalInSecs 0.2`

Unfortunately, `Start-Sleep -Seconds` takes an int as a parameter, so `0.2` gets casted to int and rounded to `0`. If you want sub-second resolution of sleep, you have to use the `-Milliseconds` parameter.

CIScripts/Test/Tests/vRouterAgentService.Tests.ps1
```diff
@@ -40,7 +44,9 @@ Describe "vRouter Agent service" {
 
     Context "given vRouter Forwarding Extension is NOT running" {
         It "crashes" {
-            Assert-AgentProcessCrashed -Session $Session -After $TimeBeforeCrash
+            {{ Test-HasAgentProcessCrashed -Session $Session -After $BeforeCrash } `
+                | Retry-UntilTrue -IntervalInSecs 1 -TimeoutInSecs 60 } `
+                | Should Not Throw
```
krdln: Although I prefer not using `Should Not Throw` here at all, it seems I'm in a minority.

Anyway, my previous comments about nested blocks applies, please use more obvious indentation or, preferably, pass the script block as explicit parameter to `Should`.

krdln (changes requested):

CIScripts/Test/TestConfigurationUtils.ps1
krdln: (this code is already removed, but if it wasn't, I'd suggest to just get rid of this variable, it's not necessary here)

krdln:
@m-kostrzewa 
> something similar to _gomega's_ `Eventually` and `Consistently`

That's why I couldn't find anything about "Pester Eventually"!
Anyway, I think it's a good idea. Syntaxwise though, I think the whole `Get ... | Should` would need to be passed a single `ScriptBlock` to `Eventually`/`Consistently`. (Actually, the current `Retry-Command` is an implementation of `Eventually`, am I right?)

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -203,69 +205,71 @@ function Disable-AgentService {
...
 
-        if ($Res) {
-            return
-        }
+function Assert-IsAgentServiceEnabled {
+    Param ([Parameter(Mandatory = $true)] [PSSessionT] $Session)
+    $Status = Retry-Command -IntervalInSecs $TIME_BETWEEN_AGENT_CHECKS_IN_SECONDS `
+                            -TimeoutInSecs $MAX_WAIT_TIME_FOR_AGENT_IN_SECONDS `
+                            { Get-AgentServiceStatus -Session $Session }
+    if ($Status -eq "Running") {
```
krdln: I think it has different semantics now vs. the original – it won't work in a case when agent service is "Stopped" for a moment.

CIScripts/Test/PesterHelpers/PesterHelpers.ps1
```diff
@@ -0,0 +1,67 @@
...
+    ScriptBlock containing a Pester assertion.
+    .PARAMETER Interval
+    Interval (in seconds) between retries.
+    .PARAMETER Duration
+    Maximum to spend between retries. After this time, Eventually fails.
+    #>
+    Param (
+        [Parameter(Mandatory=$true)] [ScriptBlock] $ScriptBlock,
+        [Parameter(Mandatory=$false)] [int] $Interval = 1,
+        [Parameter(Mandatory=$false)] [int] $Duration = 3
```
krdln: I think the default duration should be closer to inifinity (a minute or few). Or perhaps, make it mandatory?
Also, in my opinion, it'd be better to call this parameter `Timeout`.

CIScripts/Test/PesterHelpers/PesterHelpers.ps1
```diff
@@ -0,0 +1,67 @@
...
+    ScriptBlock containing a Pester assertion.
+    .PARAMETER Interval
+    Interval (in seconds) between retries.
+    .PARAMETER Duration
+    Total time to spend between retries.
+    #>
+    Param (
+        [Parameter(Mandatory=$true)] [ScriptBlock] $ScriptBlock,
+        [Parameter(Mandatory=$false)] [int] $Interval = 1,
+        [Parameter(Mandatory=$false)] [int] $Duration = 3
```
krdln: Perhaps make `Duration` mandatory?

CIScripts/Test/PesterHelpers/PesterHelpers.ps1
```diff
@@ -0,0 +1,67 @@
...
+    )
+    if ($Duration -lt $Interval) {
+        throw "Duration must be longer than interval"
+    }
+    $NumRetries = $Duration / $Interval
+    $Success = $false
+    foreach($i in 1..$NumRetries) {
```
krdln: Space, please :)

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -335,6 +339,12 @@ function Wait-RemoteInterfaceIP {
     }
 }
 
+function Initialize-DriverAndExtension {
+    Param ([Parameter(Mandatory = $true)] [PSSessionT] $Session,
+           [Parameter(Mandatory = $true)] [TestConfiguration] $TestConfiguration)
+    Initialize-TestConfiguration -Session $Session -TestConfiguration $TestConfiguration
```
krdln: Perhaps with `-NoNetwork $true`?

CIScripts/Test/Tests/vRouterAgentService.Tests.ps1
```diff
@@ -0,0 +1,119 @@
...
+
+        BeforeEach {
+            Disable-VRouterExtension -Session $Session `
+                -AdapterName $TestConf.AdapterName `
+                -VMSwitchName $TestConf.VMSwitchName `
+                -ForwardingExtensionName $TestConf.ForwardingExtensionName
+            Test-IsVRouterExtensionEnabled -Session $Session `
+                -VMSwitchName $TestConf.VMSwitchName `
+                -ForwardingExtensionName $TestConf.ForwardingExtensionName `
+                | Should Be $false
```
krdln: Is using `Should` outside `It` a good practice?

CIScripts/Test/Tests/vRouterAgentService.Tests.ps1
```diff
@@ -0,0 +1,119 @@
...
+            Enable-AgentService -Session $Session
+        }
+    }
+    
+    Context "vRouter Forwarding Extension was disabled while Agent was running" {
+        It "crashes" {
+            Eventually {
+                Read-SyslogForAgentCrash -Session $Session -After $BeforeCrash `
+                    | Should Not BeNullOrEmpty
+            }
```
krdln: It's using timeout = 3 here, which may be not enough.

krdln (changes requested):

CIScripts/Test/PesterHelpers/PesterHelpers.ps1
m-kostrzewa: I'm not sure, it can be unnecessary boilerplate.

krdln: This function is not used very often (compared to eventually), so I'm not sure about providing default.

CIScripts/Test/PesterHelpers/PesterHelpers.ps1
m-kostrzewa: It should be as short as possible by default, so that the tests are as fast as possible. Parameter names are from Gomega.
Default values there are:
`[Eventually] The default value for the timeout is 1 second and the default value for the polling interval is 10 milliseconds`
`Consistently will monitor every 10 milliseconds for up to 100 milliseconds`

I decided not to make the defaults as low as those, since they are designed for single-program goroutine testing. 3 seconds seem fine for network stuff and services enabling/disabling.

CIScripts/Test/Tests/vRouterAgentService.Tests.ps1
m-kostrzewa: Every time in `vRouter Forwarding Extension was disabled while Agent was running, it crashes` this timeout was enough. `given vRouter Forwarding Extension is NOT running, it crashes` however needed a longer timeout. I would leave it short for now, increase it only if it's needed.

CIScripts/Test/Tests/vRouterAgentService.Tests.ps1
m-kostrzewa: This was copied over from old powershell script. I'll remove it.

CIScripts/Common/Retry-Command.ps1
```diff
@@ -0,0 +1,37 @@
...
+            return $ReturnVal
+        } catch {
+            Write-Host -NoNewline "."
+            $HasPrintedDots = $true
+            Start-Sleep -s $IntervalInSecs
+        }
+    }
+    Write-Host ""
+    throw "Retry-Command: command timed out."
+}
```
kaweue: New line ?

CIScripts/Test/PesterHelpers/PesterHelpers.ps1
m-kostrzewa: Why would you like to make `Duration` mandatory?

CIScripts/Test/TestConfigurationUtils.ps1
m-kostrzewa: You mean that

`throws -> throws -> throws -> Running`
`throws -> throws -> Stopped -> Running`

in the second case, `Retry-Command` will return `Stopped` prematurely? And old implementation would handle it properly?

Makes sense. Will fix

CIScripts/Common/Retry-Command.ps1
```diff
@@ -0,0 +1,37 @@
+function Retry-Command {
```
kaweue: Maybe Poll-Command ?

CIScripts/Test/Tests/vRouterAgentService.Tests.ps1
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        BeforeEach {
+            Test-IsVRouterExtensionEnabled -Session $Session `
```
kaweue: Is this the same code as in line 59 ?

kaweue (changes requested):

m-kostrzewa:
You can see the output here http://ci.wincontrail.codilime.com:8080/job/Wincontrail/job/test-integration-debug/577/console

CIScripts/Test/Tests/vRouterAgentService.Tests.ps1
```diff
@@ -116,4 +100,10 @@ Describe "vRouter Agent service" {
         Clear-TestConfiguration -Session $Session -TestConfiguration $TestConf
         Uninstall-Agent -Session $Session
     }
+
+    BeforeAll {
+        if (-not $WithWriteHost) {
+            Mock Write-Host { return }
```
kaweue: I'm not sure if we should by default suppress logs from tests.

kaweue (changes requested):

m-kostrzewa:
Tests in Codi CI fail after the merge, which was expected.

CIScripts/Common/Exceptions.ps1
```diff
@@ -0,0 +1,4 @@
+class CITimeoutException : System.Exception {
+    CITimeoutException([string] $msg) : base($msg) {}
+    CITimeoutException([string] $msg, [System.Exception] $inner) : base($msg, $inner) {}
```
krdln, line 3: TODO

CIScripts/Test/PesterHelpers/PesterHelpers.ps1
krdln: We've talked about whether `Duration` should be mandatory and we agree that it should, because if somebody has to use `Eventually`, they need to be doing it consciously and specify timeout explicitely.

CIScripts/Common/Invoke-UntilSucceeds.ps1
```diff
@@ -0,0 +1,56 @@
...
+    .PARAMETER Interval
+    Interval (in seconds) between retries.
+    .PARAMETER Duration
+    Timeout (in seconds).
+    #>
+    Param (
+        [Parameter(Mandatory=$true,
+                   ValueFromPipeline=$true)] [ScriptBlock] $ScriptBlock,
+        [Parameter(Mandatory=$false)] [int] $Interval = 1,
+        [Parameter(Mandatory=$false)] [int] $Duration = 3
```
krdln: `Duration` should be mandatory, because of the same reasoning as in `Eventually` below.

CIScripts/Common/Invoke-UntilSucceeds.ps1
```diff
@@ -0,0 +1,56 @@
...
+                throw New-Object -TypeName CITimeoutException("Did not evaluate to True." + 
+                    "Last return value encountered was: $ReturnVal.")
+            }
+        } catch {
+            $HasPrintedDots = $true
+            Write-Host -NoNewline "."
+            $LastException = $_.Exception
+            Start-Sleep -s $Interval
+        }
+    }
```
krdln: Please use `do-while` here and in other similar functions, so we can guarantee at least one iteration of the loop.

CIScripts/Common/Invoke-UntilSucceeds.ps1
```diff
@@ -0,0 +1,56 @@
+. $PSScriptRoot\Aliases.ps1
+. $PSScriptRoot\Exceptions.ps1
+
+function Invoke-UntilSucceeds {
+    <#
+    .SYNOPSIS
+    Utility wrapper for making sure that the assert is consistently true. It works by retrying
+    the assert every Interval seconds, up to Duration.
```
krdln: This description doesn't match the function name.

CIScripts/Common/Invoke-UntilSucceeds.ps1
```diff
@@ -0,0 +1,56 @@
...
+    if ($Interval -eq 0) {
+        throw "Interval must not be equal to zero"
+    }
+    $HasPrintedDots = $false
+    $StartTime = Get-Date
+    while (((Get-Date) - $StartTime).Seconds -lt $Duration) {
+        try {
+            $ReturnVal = & $ScriptBlock
+            if ($ReturnVal) {
+                break
```
krdln: This contract is not obvious - what does it mean for the command to "succeed"? Is returning `$false` a failure, or just successful return of `$false`?

CIScripts/Common/Invoke-UntilSucceeds.ps1
krdln: Also, this function tries to do two things at once - checking return value and looking at exceptions. Perhaps split this functionality or update the description and function name?

CIScripts/Common/Invoke-UntilSucceeds.ps1
```diff
@@ -0,0 +1,56 @@
...
+        } catch {
+            $HasPrintedDots = $true
+            Write-Host -NoNewline "."
+            $LastException = $_.Exception
+            Start-Sleep -s $Interval
+        }
+    }
+    if ($HasPrintedDots) {
+        Write-Host ""
+    }
```
krdln: We think that printing dots is not a responsibility of this function. If somebody wants to print them, they can add `Write-Host` to script block.

CIScripts/Test/TestRunner.ps1
```diff
@@ -30,27 +28,41 @@ function Run-TestScenarios {
...
+        # Test-MultipleSubnetsSupport -Session $Sessions[0] -TestConfiguration $TestConfiguration
+        # Test-DockerDriverMultiTenancy -Session $Sessions[0] -TestConfiguration $TestConfiguration
+        # Test-WindowsLinuxIntegration -Session $Sessions[0] -TestConfiguration (Get-TestConfigurationWindowsLinux)
+        # Test-Pkt0PipeImplementation -Session $Sessions[0] -TestConfiguration $TestConfiguration
+        
+        # if($Env:RUN_DRIVER_TESTS -eq "1") {
+        #     Test-DockerDriver -Session $Sessions[0] -TestConfiguration $TestConfiguration
+        # }
     })
+    if ($Env:RUN_PESTER_TESTS -eq $true) {
```
krdln: What happens when there's no such env?
Also, we'd prefer to reverse the logic here -- by default the tests should run.

CIScripts/Test/TestRunner.ps1
```diff
@@ -30,27 +28,41 @@ function Run-TestScenarios {
...
+        $TotalFailedCount = 0
+
+        $TestPaths = Get-ChildItem -Recurse -Filter "*.Tests.ps1"
+        Foreach ($TestPath in $TestPaths) {
+            $Results = Invoke-Pester -PassThru -Script @{Path=$TestPath.FullName; `
+                    Parameters=@{TestbedAddr=$Sessions[0].ComputerName}; Arguments=@()}
```
krdln: Please initialize this dictionary in separate variable with nice indentation (PS, you don't need backtick on the end of line when inside `{}`)

CIScripts/Test/Tests/vRouterAgentService.Tests.ps1
```diff
@@ -0,0 +1,102 @@
...
+    [Parameter(Mandatory=$true)] [string] $TestbedAddr
+)
+
+. $PSScriptRoot\..\Utils\CommonTestCode.ps1
+. $PSScriptRoot\..\TestConfigurationUtils.ps1
+. $PSScriptRoot\..\..\Common\Aliases.ps1
+. $PSScriptRoot\..\..\Common\VMUtils.ps1
+. $PSScriptRoot\..\PesterHelpers\PesterHelpers.ps1
+
+. $PSScriptRoot\..\GetTestConfigurationCodiLegacy.ps1
```
krdln, line 11: We should either use the file from environment (at least temporarily) or merge those files into one.

krdln (changes requested):
MK + PM

CIScripts/Test/Tests/vRouterAgentService.Tests.ps1
```diff
@@ -0,0 +1,102 @@
...
+
+    BeforeEach {
+        Initialize-DriverAndExtension -Session $Session -TestConfiguration $TestConf
+        Install-Agent -Session $Session
+        New-AgentConfigFile -Session $Session -TestConfiguration $TestConf
+    }
+
+    AfterEach {
+        Clear-TestConfiguration -Session $Session -TestConfiguration $TestConf
+        Uninstall-Agent -Session $Session
```
krdln: Does uninstalling agent guarantee that it's disabled? If not, we should do it explicitely

CIScripts/Common/Invoke-UntilSucceeds.ps1
m-kostrzewa: This function was designed to check return val and catch exceptions. `Invoke-UntilSucceeds` was a name that we came up with @PawelMarchewkaCodiLime. Updated the description though.

CIScripts/Common/Exceptions.ps1
m-kostrzewa: :question: :question: :question: :question:

CIScripts/Common/Invoke-UntilSucceeds.ps1
m-kostrzewa: What do you mean it's not obvious, there are test cases that cover it pretty extensively?

CIScripts/Common/Invoke-UntilSucceeds.ps1
m-kostrzewa: plus added tests

CIScripts/Test/Tests/vRouterAgentService.Tests.ps1
m-kostrzewa: I don't know what you mean. I am not going to expand this PR further to clean up the config file mess if that's what you guys want :smile:

CIScripts/Test/PesterHelpers/PesterHelpers.ps1
m-kostrzewa: I'm worried this will lead to arbitrarily large timeouts. IMO people would think that the default value is "good enough" and they don't need to change it (by increasing it "just in case"). But ok I'll make it mandatory.

I'm thinking of adding a script analyzer rule to check for this in another PR.

m-kostrzewa:
Didn't run it in legacy CI anymore, here's test output from my machine: https://hastebin.com/nipugemeyo.vbs

kaweue (approved):

CIScripts/Common/Invoke-UntilSucceeds.ps1
```diff
@@ -4,10 +4,12 @@
 function Invoke-UntilSucceeds {
     <#
     .SYNOPSIS
-    Utility wrapper for making sure that the assert is consistently true. It works by retrying
-    the assert every Interval seconds, up to Duration.
+    Repeatedly calls a script block until its return value evaluates to true. Subsequent calls
```
krdln: ~~This description doesn't match the code – it suggests that on falsy result it will retry the command, but actually it throws on the first falsy result.
I'd just remove the `"Did not evaluate to True."` check and logic, leaving it up to the caller.~~

Ok, it's catched right afterwards. Sorry for brainfart then

krdln (changes requested):

CIScripts/Common/Invoke-UntilSucceeds.ps1
krdln: I mean that command could succeed and yet return `$false`, but now we interpret the `$false`/`$null` or `0` unconditionally as a failure (I know, it's documented, but it's not obvious definition of `Succeed`). I'll accept this function as is, but I still think it does two things at once (checking exceptions ∧ checking value), which violates single responsibility principle. I think that the following design would be more elegant:

Case 1, user considers `0` as valid result:
```
$Count = Retry-UntilSucceeds { Count-Things }
```

Case 2, User considers `$null` as an error:
```
$TheThing = Retry-UntilSucceeds {
    Get-Thing | ThrowOnFalse
}
```

(where `ThrowOnFalse` is helper function – an identity function that additionally performs the check)

krdln:
~~pls rebase~~

krdln (approved):

CIScripts/Common/Invoke-UntilSucceeds.ps1
m-kostrzewa: I disagree that it violates SRP. "This function invokes the command until it evaluates to true". It reduces a lot of boilerplate all over the code. It's an "if" on steroids.

Case 1: valid, but this function isn't about counting things but retrying one command.

Case 2: $null is treated as an error, and retries will continue.

PR #60, opened 2018-01-31
Deploy artifacts to provisioned Testbeds

jablonskim:


CIScripts/Build.ps1
```diff
@@ -2,6 +2,7 @@
 
 . $PSScriptRoot\Common\Init.ps1
 . $PSScriptRoot\Common\Job.ps1
+. $PSScriptRoot\Common\Functions.ps1
```
kaweue: We can use different name like: projects.ps1.

CIScripts/Common/VMUtils.ps1
```diff
@@ -40,5 +40,5 @@ function New-RemoteSessionsToTestbeds {
 }
 
 function Get-TestbedHostnamesFromEnv {
-    return $Env:TESTBED_HOSTNAMES.Split(",")
+    return $Env:TESTBED_ADDRESSES.Split(",")
```
kaweue: Maybe Get-TestbedAddressesFromEnv

kaweue (changes requested):
JI + PM

jacek-izykowski (dismissed):
PM, JI

kaweue (dismissed):

ansible/roles/tester/tasks/main.yml
```diff
@@ -25,3 +25,10 @@
     name: VMWare.PowerCLI
     state: present
 
+- name: Check TrustedHosts value
+  win_shell: (Get-Item WSMan:\localhost\Client\TrustedHosts).Value
+  register: trusted_hosts
+
+- name: Set Testbeds as TrustedHosts
+  win_shell: "Set-Item WSMan:\\localhost\\Client\\TrustedHosts -Value \"{{ (((trusted_hosts.stdout | trim()).split(',') | reject('equalto', '') | list()) + ['172.17.0.*']) | join(',') }}\" -Force"
```
krdln: You can use the [`>` style syntax](http://docs.ansible.com/ansible/latest/YAMLSyntax.html) next time, so you don't have to do all these escapes. I think it would look like:

```yaml
  win_shell: >
    Set-Item WSMan:\localhost\Client\TrustedHosts
    -Value "{{ (((trusted_hosts.stdout | trim()).split(',') | reject('equalto', '') | list()) + ['172.17.0.*']) | join(',') }}"
    -Force
```

Also, https://stackoverflow.com/a/21699210

Perhaps you can also move computing the value out of the string to make things a little bit cleaner, by using `set_fact` or `with_items` (I'm fine with the status quo though)

jacek-izykowski (approved):

PR #61, opened 2018-01-31
Make our code compliant to durable-task-plugin 1.18

mclapinski:


CIScripts/Common/Invoke-NativeCommand.ps1
```diff
@@ -19,11 +19,11 @@ function Invoke-NativeCommand {
     #
     # Note: The command has to return 0 exitcode to be considered successful.
 
-    $Global:LastExitCode = $null
-
     & {
+        # Since we're redirecting stderr to stdout we shouldn't have to set ErrorActionPreference
+        # but because of a bug in Powershell we have to
```
m-kostrzewa: Could you link to some existing bugs about this issue here?

m-kostrzewa:
:clap:

CIScripts/Common/Invoke-NativeCommand.ps1
```diff
@@ -33,6 +33,4 @@ function Invoke-NativeCommand {
     if ($AllowNonZero) {
         Write-Output $LastExitCode
     }
-
-    $Global:LastExitCode = $null
```
krdln: I guess this is not needed, because now the `$Error` [is checked before](https://github.com/jenkinsci/durable-task-plugin/blob/4a06f3cc8eb9053c0bf5597a06b69163c005c3b0/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java#L103) `$LastExitCode`? Is my understanding correct?

Anyway, I'd keep this line, so we don't lose compatibility with currently used version of durable task plugin.
(also, it's needed, so next time `$LastExitCode -ne 0` fails on "Command not found" errors and the like)

mclapinski: It passed CI, so it is compatible with current version of durable-task-plugin

krdln: You're right, it passed CI, but unfortunatelly it detects less errors. With this line removed the following errors won't be catched:
1. Any error thrown after successful `Invoke-NativeCommand` (only with current version of durable task plugin)
2. Errors like "Command not found" (also with new version)

krdln (approved):
Thanks for adding comments with explanations of each hack!

MK (let's wait for pair review)

CIScripts/Common/Invoke-NativeCommand.ps1
```diff
@@ -1,7 +1,7 @@
 function Invoke-NativeCommand {
     Param (
-        [Parameter(Mandatory = $true)] [ScriptBlock] $ScriptBlock,
-        [Parameter(Mandatory = $false)] [Bool] $AllowNonZero = $false
+        [Parameter(Mandatory = $false)] [Bool] $AllowNonZero = $false,
```
krdln, line 3: We can use flag here instead of parameter (didn't know about it when I've written it in the first place):
```
function Foo
{
  param(
    [switch] $AllowNonZero 
  )

  if ($AllowNonZero) { ... }
}
```

CIScripts/Build/BuildFunctions.ps1
```diff
@@ -270,12 +268,11 @@ function Run-Test {
...
     $Res = Invoke-Command -ScriptBlock {
-        $ErrorActionPreference = "SilentlyContinue"
-        $TestOutput = Invoke-NativeCommand -AllowNonZero $true -ScriptBlock {
+        $NativeCommandReturn = Invoke-NativeCommand -AllowNonZero $true -ScriptBlock {
             Invoke-Expression $TestExecutable
         }
-        $ExitCode = $TestOutput[-1]
-        $TestOutput.ForEach({ Write-Host $_ })
+        $ExitCode = $NativeCommandReturn[-1]
+        $TestOutput = $NativeCommandReturn[0..($NativeCommandReturn.Length-2)]
```
m-kostrzewa, line 275: This isn't really clear, but it's not really part of this PR, so fix it if you want.

Maybe we should add:
1. `CaptureOutput` flag to `Invoke-NativeCommand`
2. Return a structure from `Invoke-NativeCommand` with two fields: `ExitCode` and `TestOutput`
3. If `CaptureOutput` is `$false`, then `TestOutput` would be `$null`

krdln: > If CaptureOutput is $false, then TestOutput would be $null

And it should be live-streamed through `Write-Host`

m-kostrzewa (approved):

krdln (approved):
MK + MK

CIScripts/Build/BuildFunctions.ps1
mclapinski: I tried to do this, it's actually pretty hard because PS never does what you expect it to do, so I won't do this.

PR #67, opened 2018-02-05
Execute post stage on any linux node instead of master

krdln:
When merged into production, should fix deadlock caused by running both proxy jobs and post-stages on `master`.

(Ideally, the proxy jobs should not be executed on master)


kaweue (approved):

kaweue:
MJ + PM

PR #70, opened 2018-02-06
Deploy VM from template using vSphere API

sodar:


ansible/vmware_common.py
```diff
@@ -0,0 +1,76 @@
...
+
+    hosts = sorted(hosts, key=lambda h: h.name)
+    dest_host_index = build_id % len(hosts)
+    dest_host = hosts[dest_host_index]
+
+    return dest_host
+
+
+def select_destination_host_datastore_by_free_space(host):
+    datastores = [d for d in host.datastore if 'ssd' in d.name]
```
m-kostrzewa: Is hardcoded string `'ssd'` dependent on actual infrastructure names? If so, we should probably make it a param or something.

ansible/vmware_common.py
```diff
@@ -0,0 +1,76 @@
...
+        raise Exception("Couldn't find the VirtualMachine with the provided name "
+                        "'{}'".format(vm_name))
+    return vm
+
+
+def find_vm_folder(conn, folder_path):
+    inventory_path = os.path.join(conn.datacenter.name, 'vm', folder_path)
+    folder = conn.content.searchIndex.FindByInventoryPath(inventory_path)
+    if not folder:
+        raise Exception("Couldn't find the Folder with the provided name "
```
m-kostrzewa: In all cases in the code, `Exception` class is too generic. I would either choose a more specific subclass from here https://docs.python.org/3/library/exceptions.html (in this case, maybe `FileNotFoundException` would work), or create my own subclass, like `VMWareException` / `ResourceNotFoundException`...

ansible/deploy_vm_from_template.py
```diff
@@ -0,0 +1,237 @@
...
+            return []
+
+    device_list = []
+    for device in vm.config.hardware.device:
+        if isinstance(device, vim.vm.device.VirtualPCNet32) or \
+           isinstance(device, vim.vm.device.VirtualVmxnet2) or \
+           isinstance(device, vim.vm.device.VirtualVmxnet3) or \
+           isinstance(device, vim.vm.device.VirtualE1000) or \
+           isinstance(device, vim.vm.device.VirtualE1000e) or \
+           isinstance(device, vim.vm.device.VirtualSriovEthernetCard):
```
m-kostrzewa: All these classes extend `VirtualEthernetCard` (https://www.vmware.com/support/developer/converter-sdk/conv55_apireference/vim.vm.device.VirtualEthernetCard.html) - maybe we can only check 
`isinstance(device, vim.vm.device.VirtualEthernetCard)`?

ansible/deploy_vm_from_template.py
```diff
@@ -0,0 +1,237 @@
...
+    nic.device.backing.port = dvs_port_connection
+
+    return nic
+
+
+def deploy_vm_from_template(service_instance, datacenter_name, cluster_name, template_name,
+                            destination_folder, build_id, mgmt_network_name,
+                            data_network_name, data_ip_address, data_netmask,
+                            vm_name, vm_password, vm_org='Contrail',
+                            vm_administrator='Administrator'):
```
m-kostrzewa: This smells of too many params

ansible/deploy_vm_from_template.py
```diff
@@ -0,0 +1,237 @@
...
+
+def get_args():
+    parser = argparse.ArgumentParser(description='Arguments for talking to vCenter')
+
+    parser.add_argument('--build-id',
+                        required=True,
+                        type=int,
+                        action='store',
+                        help='vSphere service to connect to')
+
```
m-kostrzewa: [Personal taste], but these newlines seem unnecessary, especially since we have big indents. Personnaly I would remove them. I can't find what pep8 says about this.

ansible/deploy_vm_from_template.py
```diff
@@ -0,0 +1,237 @@
...
+
+def get_args():
+    parser = argparse.ArgumentParser(description='Arguments for talking to vCenter')
+
+    parser.add_argument('--build-id',
+                        required=True,
+                        type=int,
+                        action='store',
+                        help='vSphere service to connect to')
```
m-kostrzewa: Help text `vSphere service to connect to` is confusing when the param is called `build-id`. Could you expand the help a little bit?

ansible/deploy_vm_from_template.py
```diff
@@ -0,0 +1,237 @@
...
+           isinstance(device, vim.vm.device.VirtualVmxnet3) or \
+           isinstance(device, vim.vm.device.VirtualE1000) or \
+           isinstance(device, vim.vm.device.VirtualE1000e) or \
+           isinstance(device, vim.vm.device.VirtualSriovEthernetCard):
+            device_list.append(device)
+
+    return device_list
+
+
+def get_nic_with_updated_network(content, nic_device, network_name):
```
m-kostrzewa: Looking at the context in which this function is used, maybe calling it `get_nic_update_network_operation` would be better?

ansible/deploy_vm_from_template.py
m-kostrzewa: Hmm how would I pursue refactoring this...
* arguments with `vm` prefix look like some class.

m-kostrzewa: So looking through this function, I've drawn this diagram. It shows whether a variable (or parameter) is required to create another variable. It forms a DAG. We can look for clusters to guide our refactoring.
dario1

One such clustering could look like this:
dario2

From this we can see classes with constructors / function signatures like:
`connection(service_instance, datacenter_name, cluster_name)`
`config_spec(connection, template)`
`relocation_spec(connection, build_id)`
`customization_spec(identity, data_ip_address, data_network)`
`identity(vm_name, vm_password, vm_org, vm_administrator)`
`data_guest_map(data_ip_address, data_netmask)`
`clone_spec(config_spec, relocation_spec, customizatoin_spec)`
`task(clone_spec, service_instance, destination_folder, vm_name)`

Note: of course there are dozens of ways to cluster that graph. It is just a helper.

ansible/vmware_common.py
```diff
@@ -0,0 +1,76 @@
...
+
+def find_vm(conn, vm_name):
+    vm = get_obj(conn.content, [vim.VirtualMachine], vm_name, conn.datacenter.vmFolder)
+    if not vm:
+        raise Exception("Couldn't find the VirtualMachine with the provided name "
+                        "'{}'".format(vm_name))
+    return vm
+
+
+def find_vm_folder(conn, folder_path):
```
m-kostrzewa: Are we actually finding the folder or getting it? I know we call `FindByInventoryPath` below but it looks like `folder_path` is known? If so, maybe renaming this function to `get_vm_folder` would be better.

ansible/vmware_common.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def select_destination_host_using_id(conn, build_id):
+    hosts = conn.cluster.host
+
+    hosts = sorted(hosts, key=lambda h: h.name)
+    dest_host_index = build_id % len(hosts)
+    dest_host = hosts[dest_host_index]
+
+    return dest_host
```
m-kostrzewa: [Personal taste] I don't like these empty lines.

ansible/vmware_common.py
```diff
@@ -0,0 +1,76 @@
...
+def select_destination_host_datastore_by_free_space(host):
+    datastores = [d for d in host.datastore if 'ssd' in d.name]
+    datastores = sorted(datastores, key=lambda d: d.summary.freeSpace)
+    dest_datastore = datastores[-1]
+    return dest_datastore
+
+
+def wait_for_task(conn, task):
+    while task.info.state in ['queued', 'running']:
+        time.sleep(1)
```
m-kostrzewa: I would be scared running this in prod with no timeout. I know that in theory Jenkins should just kill this job if it sleeps for too long, but we never know when someone disables/misconfigures timeouts on Jenkins or something. I think we should be defensive here.
Also someone could run those playbooks manually, not from Jenkins - what happens then?

ansible/deploy_vm_from_template.py
m-kostrzewa: BTW similar refactoring method could be used in other parts of the code, like `get_nic_with_updated_network` function.

m-kostrzewa (dismissed):
Good effort overall, gj.

Aside from the comments, I would also try to have more logging output. I would recommend using `import logging`.

Also, we need more unit tests and maybe even integration tests that would run this in CI. We do expect to change this code in the future, after all.

ansible/deploy_vm_from_template.py
```diff
@@ -0,0 +1,237 @@
...
+
+    parser.add_argument('--vm-password',
+                        required=False,
+                        default=None,
+                        action='store',
+                        help='Password to set for Administrator on Windows')
+
+    args = parser.parse_args()
+
+    if not args.password:
```
kaweue: I don't think we need that. Do we?  As I understand we are going to use this script in CI.

ansible/deploy_vm_from_template.py
```diff
@@ -0,0 +1,237 @@
...
+import atexit
+import argparse
+import getpass
+from pyVim.connect import SmartConnectNoSSL, Disconnect
+from pyVmomi import vim, vmodl
+
+import vmware_common as vm
+
+
+def get_args():
```
kaweue: There is so many required arguments, maybe this configuration could be loaded from some file, like JSON file or even YAML....

kaweue: Name of the file could be passed as an argument.

ansible/deploy_vm_from_template.py
```diff
@@ -0,0 +1,237 @@
...
+
+    device_list = []
+    for device in vm.config.hardware.device:
+        if isinstance(device, vim.vm.device.VirtualPCNet32) or \
+           isinstance(device, vim.vm.device.VirtualVmxnet2) or \
+           isinstance(device, vim.vm.device.VirtualVmxnet3) or \
+           isinstance(device, vim.vm.device.VirtualE1000) or \
+           isinstance(device, vim.vm.device.VirtualE1000e) or \
+           isinstance(device, vim.vm.device.VirtualSriovEthernetCard):
+            device_list.append(device)
```
kaweue: Maybe just create and return list here

ansible/deploy_vm_from_template.py
```diff
@@ -0,0 +1,237 @@
...
+    dvs_port_connection = vim.dvs.PortConnection()
+    dvs_port_connection.portgroupKey = pg_obj.key
+    dvs_port_connection.switchUuid = pg_obj.config.distributedVirtualSwitch.uuid
+    nic.device.backing = vim.vm.device.VirtualEthernetCard.DistributedVirtualPortBackingInfo()
+    nic.device.backing.port = dvs_port_connection
+
+    return nic
+
+
+def deploy_vm_from_template(service_instance, datacenter_name, cluster_name, template_name,
```
kaweue: I have to agree this functions is a bit longish

ansible/deploy_vm_from_template.py
```diff
@@ -0,0 +1,237 @@
...
+    host = vm.select_destination_host_using_id(connection, build_id)
+    datastore = vm.select_destination_host_datastore_by_free_space(host)
+
+    # Create `VirtualMachineConfigSpec` - reconfigure networks
+    config_spec = vim.vm.ConfigSpec()
+    config_spec.deviceChange = []
+
+    current_network_devices = get_vm_network_interfaces(template)
+
+    eth0 = get_nic_with_updated_network(connection.content, current_network_devices[0], mgmt_network_name)
```
kaweue: Is it really eth0 or just primary interface ?

ansible/deploy_vm_from_template.py
```diff
@@ -0,0 +1,237 @@
...
+                                data_network_name=args.data_network,
+                                data_ip_address=args.data_ip_address,
+                                data_netmask=args.data_netmask,
+                                vm_name=args.name,
+                                vm_password=args.vm_password)
+    except Exception as err:
+        # TODO: Handle errors
+        raise
+
+    Disconnect(service_instance)
```
kaweue: Shouldn't we Disconnect anyway even in case of exception in some cases (like exception in deploy_vm_from_template)?

ansible/deploy_vm_from_template.py
```diff
@@ -0,0 +1,237 @@
...
+    vm.wait_for_task(service_instance, task)
+
+
+def main():
+    args = get_args()
+
+    try:
+        build_id = int(args.build_id)
+
+        service_instance = SmartConnectNoSSL(host=args.host, user=args.user, pwd=args.password)
```
kaweue: Could we use somthing like this ```with SmartConnection(host=args.host, user=args.user, pwd=args.password) as service ``` without using try and except ??

ansible/deploy_vm_from_template.py
```diff
@@ -0,0 +1,237 @@
...
+    # Create `VirtualMachineCloneSpec` consumed by clone template operation
+    clone_spec = vim.vm.CloneSpec()
+    clone_spec.powerOn = False
+    clone_spec.template = False
+    clone_spec.config = config_spec
+    clone_spec.customization = customization_spec
+    clone_spec.location = relocation_spec
+
+    task = template.Clone(name=vm_name, folder=folder, spec=clone_spec)
+    vm.wait_for_task(service_instance, task)
```
kaweue: Why can't we use method WaitForTask from task ?

ansible/vmware_common.py
```diff
@@ -0,0 +1,76 @@
...
+
+def select_destination_host_datastore_by_free_space(host):
+    datastores = [d for d in host.datastore if 'ssd' in d.name]
+    datastores = sorted(datastores, key=lambda d: d.summary.freeSpace)
+    dest_datastore = datastores[-1]
+    return dest_datastore
+
+
+def wait_for_task(conn, task):
+    while task.info.state in ['queued', 'running']:
```
kaweue: Are those values documented somewhere ?

ansible/deploy_vm_from_template.py
kaweue: Could we just make a loop for all specified interfaces ?

ansible/deploy_vm_from_template.py
```diff
@@ -0,0 +1,237 @@
...
+                        required=True,
+                        action='store',
+                        help='Folder in which the cloned VM will be placed')
+
+    parser.add_argument('--name',
+                        required=True,
+                        action='store',
+                        help='Name which will be given to the cloned VM')
+
+    parser.add_argument('--mgmt-network',
```
kaweue: Maybe this code could be a bit more generic if we were able to pass list of networks with some parameters. In current solution if we would like to set up fixed ip in mgmt-network for a particular VM, we have to change that python code not only configuration.

kaweue (changes requested):

ansible/deploy_vm_from_template.py
sodar: 👍

ansible/deploy_vm_from_template.py
sodar: This help text is incorrect. It was left after copy-pasting `add_argument`. I will provide the correct message there.

ansible/files/provision_vm.py
```diff
@@ -0,0 +1,154 @@
...
+
+    parser.add_argument('--host',
+                        required=True,
+                        action='store',
+                        help='vSphere service to connect to')
+
+    parser.add_argument('--user',
+                        required=True,
+                        action='store',
+                        help='User name to use')
```
kaweue: please remove obvious comments

ansible/files/provision_vm.py
```diff
@@ -0,0 +1,154 @@
...
+        'data_netmask': args.data_netmask
+    }
+
+    config_spec = get_vm_config_spec(api, vm=template, networks=[args.mgmt_network, args.data_network])
+    customization_spec = get_vm_customization_spec(template, **customization_data)
+    relocate_spec = get_vm_relocate_spec(api.cluster, host, datastore)
+    clone_spec = get_vm_clone_spec(config_spec, customization_spec, relocate_spec)
+
+    task = template.Clone(name=name, folder=folder, spec=clone_spec)
+    api.wait_for_task(task)
```
kaweue: Please call WaitForTask on task

sodar:
recheck no bug

kaweue (changes requested):

ansible/files/vmware_api.py
```diff
@@ -0,0 +1,182 @@
...
+        datastores = [d for d in self.datacenter.datastore if 'ssd' in d.name]
+        if len(datastores) == 0:
+            return None, None
+        datastores = sorted(datastores, key=lambda d: d.info.freeSpace)
+        chosen_datastore = datastores[-1]
+        chosen_host = chosen_datastore.host[0].key
+        return chosen_host, chosen_datastore
+
+
+    def wait_for_task(self, task):
```
mclapinski: https://i.imgur.com/0oepTPd.jpg

mclapinski (changes requested):

mclapinski (approved):

kaweue (approved):

PR #73, opened 2018-02-07
Fix testbed lock race

jablonskim:


Jenkinsfile
```diff
@@ -83,50 +82,49 @@ pipeline {
             steps {
                 script {
                     try {
-                        lock(dataNetwork) {
-                            // 'Provision' stage
+                        lock('vmware-lock') {
```
kaweue: Please check if jobs take slots.

kaweue: Try to split this 'Provision & Deploy & Test' into 3 stages

kaweue (dismissed):

ansible/wait_for_folder.py
```diff
@@ -0,0 +1,101 @@
...
+    subfolder = folder.CreateFolder(args.name)
+    if not subfolder:
+        raise Exception("Couldn't create the Folder with the provided name "
+                        "'{}'".format(args.name))
+
+    return 0
+
+
+if __name__ == "__main__":
+    exit(main())
```
krdln: Perhaps we can merge this two files and use something like command argument?
Eg.
```
python vm_folder.py --datacenter foo ... wait
```
and 
```
python vm_folder.py --datacenter foo ... remove --folder bar
```

jablonskim:
recheck no bug

jablonskim:
recheck no bug

jablonskim:
recheck no bug

jablonskim:
recheck no bug

Jenkinsfile
```diff
@@ -65,100 +81,146 @@ pipeline {
...
 
-        // Variables are not supported in declarative pipeline.
-        // Possible workaround: store SpawnedTestbedVMNames in stashed file.
-        // def SpawnedTestbedVMNames = ''
-
-        // NOTE: Currently nesting multiple stages in lock directive is unsupported
-        stage('Provision & Deploy & Test') {
-            agent none
+        stage('Lock') {
+            agent { label 'ansible' }
```
sodar, line 85: Should we use `renter` instead of `ansible` here?

sodar (changes requested):

Jenkinsfile
jablonskim: Yes, but after PR with creating and deploying `renter` VM. For now we have to use `ansible`

sodar (approved):

PR #75, opened 2018-02-07
Fix error message of Invoke-NativeCommand

krdln:
It was showing totally different `ScriptBlock` than the one that failed, leading to really confusing logs.


jablonskim (approved):

jablonskim:
MJ & DS

PR #81, opened 2018-02-08
Proposed spec for Write-Log

m-kostrzewa:
[Rendered](https://github.com/codilime/contrail-windows-ci/blob/write-log/specs/Write-Log.spec.md)

### DIFF
--- a/specs/Write-Log.spec.md
+++ b/specs/Write-Log.spec.md
@@ -1,0 +1,43 @@
+# Write-Log
+
+## Problem statement
+
+Currently, `Write-Host` Powershell cmdlet is used all over test code.
+This is done because previously used `Write-Output` messes with return values of 
+nested functions.
+
+However, the side effect of using `Write-Host` is that we have no control over
+where it prints. It always prints to console - but it interferes with nice
+formating of e.g. Pester.
+
+Also, default rules of PSScriptAnalyzer recommend against using `Write-Host`
+(https://github.com/PowerShell/PSScriptAnalyzer/blob/3dc995d7a9eaf259077c469ad8468aefb8eed133/RuleDocumentation/AvoidUsingWriteHost.md)
+
+Recommendation is to use `Write-Output` or `Write-Verbose`.
+
+`Write-Output` is problematic, for the reasons stated above.
+
+`Write-Verbose` is nice, but it's hard to redirect its output to e.g. a file.
+
+## Proposal
+
+Implement a `Write-Log` function and replace every instance of `Write-Host` with it.
+
+## Requirements
+
+* By default, prints everything to a `log-detailed.txt` file.
+* Prepends the message with location (file/line/class) of where `Write-Log` was called from.
+* Can change the global filepath to the logfile.
+* Ability to specify verbosity levels: `Error`, `Warn`, `Info`, `Debug`. This just prepends the message with a tag, so that it's easier to grep.
+* Default verbosity level is `Debug`.
+* `log-detailed.txt` is rsync'ed to the log server in Jenkinsfile amongside other logs.
+
+Example fragment of `log-detailed.txt`:
+
+```
+[Error] [BuildFunctions:294]         Failed (exit code: 1).
+...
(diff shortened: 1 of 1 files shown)


specs/Write-Log.spec.md
```diff
@@ -0,0 +1,43 @@
...
+formating of e.g. Pester.
+
+Also, default rules of PSScriptAnalyzer recommend against using `Write-Host`
+(https://github.com/PowerShell/PSScriptAnalyzer/blob/3dc995d7a9eaf259077c469ad8468aefb8eed133/RuleDocumentation/AvoidUsingWriteHost.md)
+
+Recommendation is to use `Write-Output` or `Write-Verbose`.
+
+`Write-Output` is problematic, for the reasons stated above.
+
+`Write-Verbose` is nice, but it's hard to redirect its output to e.g. a file.
```
krdln, line 20: It seems that `Write-Verbose` would be perfect here. Why is it hard to redirect it to a file?

specs/Write-Log.spec.md
```diff
@@ -0,0 +1,43 @@
...
+## Proposal
+
+Implement a `Write-Log` function and replace every instance of `Write-Host` with it.
+
+## Requirements
+
+* By default, prints everything to a `log-detailed.txt` file.
+* Prepends the message with location (file/line/class) of where `Write-Log` was called from.
+* Can change the global filepath to the logfile.
+* Ability to specify verbosity levels: `Error`, `Warn`, `Info`, `Debug`. This just prepends the message with a tag, so that it's easier to grep.
```
krdln, line 31: Feels a little like reinventing the wheel, powershell already has these four outputs.

krdln:
Would it be possible to convince Pester to capture these four powershell outputs?

Also, do you want it to do only for tests or the whole pipeline?

krdln:
Question – Which output does the pester use to write its output?

mclapinski:
recheck no bug

specs/Write-Log.spec.md
m-kostrzewa: I don't think we want clients of `Write-Verbose` to specify how and where the log file should be stored. We probably want to control it globally, e.g. we don't want users to write code like:

`Write-Verbose "asdf" -Level Warn | Out-File my-logfile.txt`

but

`Write-Log -Warn "asdf"`

And somewhere in `Init.ps1` we could have `Set-LogOutputFile asdf`

m-kostrzewa: Also,` Write-Verbose` uses some pipes which might be problematic with `Invoke-NativeCommand`? @michal-clapinski

m-kostrzewa:
@krdln seems like Pester uses `Write-Host`: https://github.com/pester/Pester/blob/1520913697149f1a27b5ee026028ddf1f0076b35/Functions/Output.ps1

krdln:
@m-kostrzewa In that case, can use `Write-Verbose` and just redirect the whole Pester invocation using `4>Verbose.log`. The only problem I see is that this way various outputs (`Debug`, `Warning`, `Verbose`) would be uncorrelated in different files. Perhaps redirecting all the streams to the same file would work, though.

krdln:
recheck some bugs

m-kostrzewa:
@krdln I don't want to redirect pester to `detailed-log.txt`. I would like Pester output to stay where it is.

specs/Write-Log.spec.md
krdln: wrt. `Invoke-NativeCommand`, if it will be properly split into two variants (capturing and non capturing), then it will be simple to adapt to whichever log approach we choose.

krdln:
> @krdln I don't want to redirect pester to detailed-log.txt. I would like Pester output to stay where it is.

I don't want too, I suggested to just use `Write-Verbose` in our scripts, and then, in `Run-TestsScenarios`, add a verbose-stream redirection (`4>foo`) to the `Invoke-Pester` line. The Pester's Write-Host-output stays intact.

PR #82, opened 2018-02-08
Enable vtest

krdln:


krdln:
recheck no bug

krdln:
[works when testbeds are spawned from a new template](http://10.84.12.26:8080/job/WinContrail/job/winci-server2016-contrail-windows-ci/377/console)
(https://github.com/codilime/contrail-windows-ci/pull/92)

CIScripts/Deploy/Deployment.ps1
```diff
@@ -63,6 +63,9 @@ function Install-Artifacts {
         }
     }
 
+    Write-Host "Copying dlls"
```
m-kostrzewa, line 66: IMHO this should be `Job.Step`.

CIScripts/Test/GetTestConfigurationJuni.ps1
```diff
@@ -0,0 +1,67 @@
+function Get-TestConfiguration {
+    [TestConfiguration] @{
+        ControllerIP = "10.7.0.54";
+        ControllerRestPort = 8082;
+        ControllerHostUsername = "ubuntu";
+        ControllerHostPassword = "ubuntu";
+        AdapterName = "Ethernet1";
+        VMSwitchName = "Layered?Ethernet1";
```
m-kostrzewa: We're wondering whether this wildcard requires a comment. `Depending on what updates are installed on Windows 2016, default VMSwitchName will be Layered EthernetX or Layerd_Ethernet X`.

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -71,8 +71,16 @@ function Enable-VRouterExtension {
     Write-Host "Enabling Extension"
 
     Invoke-Command -Session $Session -ScriptBlock {
-        New-ContainerNetwork -Mode Transparent -NetworkAdapterName $Using:AdapterName -Name $Using:ContainerNetworkName | Out-Null
-        Enable-VMSwitchExtension -VMSwitchName $Using:VMSwitchName -Name $Using:ForwardingExtensionName | Out-Null
+        docker network create -d transparent --ipam-driver windows testnet
```
m-kostrzewa: 1. `AdapterName` is not specified for transparent docker network. For this reason, it will probably be created on Ethernet0. Which is probably why you experienced connection drops.

2. Why are we switching to docker network create fron New-ContainerNetwork here? docker network create requires docker to be running. If it's not, it will be enabled. So this has a side effect. New-ContainerNetwork uses HNS directly, it doesn't care if docker is even installed on the system.

3. Even if we used docker here, we don't understand the need for specifying the ipam driver. This is probably a workaround for dropped connections, but in that case, see point 1.

4. AdapterName param is not used, but see point 1.

5. ContainerNetworkName should be used in place of hardcoded `testnet`. But see point 2.

CIScripts/Test/TestRunner.ps1
```diff
@@ -44,16 +44,26 @@ function Run-TestScenarios {
...
         # Test-WindowsLinuxIntegration -Session $Sessions[0] -TestConfiguration (Get-TestConfigurationWindowsLinux)
         # Test-Pkt0PipeImplementation -Session $Sessions[0] -TestConfiguration $TestConfiguration
-        
-        # if($Env:RUN_DRIVER_TESTS -eq "1") {
-        #     Test-DockerDriver -Session $Sessions[0] -TestConfiguration $TestConfiguration
-        # }
+
+        if($Env:RUN_DRIVER_TESTS -eq "1") {
+            Test-DockerDriver -Session $Sessions[0] -TestConfiguration $TestConfiguration
+        }
```
m-kostrzewa: Please leave this commented for now - so we don't mistakenly think this is already fixed.

CIScripts/Test/TestRunner.ps1
```diff
@@ -44,16 +44,26 @@ function Run-TestScenarios {
...
+        "vRouterAgentService.Tests.ps1"
+    );
+
     $TotalFailedCount = 0
 
     $TestPaths = Get-ChildItem -Recurse -Filter "*.Tests.ps1"
     Foreach ($TestPath in $TestPaths) {
+        if ($TestPath.Name -in $TestsBlacklist) {
+            Write-Host "Skipping $($TestPath.Name)"
+            continue;
```
m-kostrzewa: Semicolon not required here

CIScripts/Test/Tests/VTestScenariosTest.ps1
```diff
@@ -7,19 +7,24 @@ function Test-VTestScenarios {
...
-            # we don't need to check the exit code because this script raises an exception on failure
-            vtest\all_tests_run.ps1 -VMSwitchName $Using:VMSwitchName -TestsFolder vtest\tests | Write-Host
+            .\vtest\all_tests_run.ps1 -VMSwitchName $Using:VMSwitchName -TestsFolder vtest\tests
 
             Pop-Location
         }
 
         Clear-TestConfiguration -Session $Session -TestConfiguration $TestConfiguration
+        Uninstall-Utils -Session $Session
+        Uninstall-Extension -Session $Session
```
m-kostrzewa: Since all_tests_run.ps1 throws an exception, this cleanup step should be in finally block

CIScripts/Test/Utils/CommonTestCode.ps1
```diff
@@ -185,3 +185,67 @@ function Ping-Container {
 
     Assert-PingSucceeded -Output $PingOutput
 }
+
+function Install-Msi {
+    Param (
+        [Switch] $Uninstall,
```
m-kostrzewa: We should have 3 functions: Install-MSI, Uninstall-MSI and Invoke-MSIExec with flags.

Install-MSI with -Unistall flag looks too weird.

m-kostrzewa: But +1 for finding this opportunity to refactor, the rest of the code that install different components is much clearer.

CIScripts/Test/Utils/CommonTestCode.ps1
```diff
@@ -185,3 +185,67 @@ function Ping-Container {
...
+
+function Install-Utils {
+    Param ([Parameter(Mandatory = $true)] [System.Management.Automation.Runspaces.PSSession] $Session)
+
+    Write-Host "Installing vRouter utilities"
+    Install-Msi -Session $Session -Path "C:\Artifacts\utils.msi"
+}
+
+function Uninstall-Utils {
+    Param ([Parameter(Mandatory = $true)] [System.Management.Automation.Runspaces.PSSession] $Session)
```
m-kostrzewa: System.Management.Automation.Runspaces.PSSession - we have an alias PSSessionT in Common/Aliases.ps1. Let's use it instead in all functions in this file.

Jenkinsfile
```diff
@@ -111,6 +111,10 @@ pipeline {
                 environment name: "DONT_CREATE_TESTBEDS", value: null
                 expression { return !currentBuild.result }
             }
+            environment {
+                TESTBED_TEMPLATE = "Template-testbed-201802130923"
```
m-kostrzewa: This is ok for this PR, but I will add to backlog that it should have a hash instead of timestamp.

Jenkinsfile
```diff
@@ -185,7 +189,7 @@ pipeline {
                     try {
                         deleteDir()
                         unstash "CIScripts"
-                        // powershell script: './CIScripts/Test.ps1'
+                        powershell script: './CIScripts/Test.ps1'
```
m-kostrzewa: While we're at it, I think we can remove TestIntegration.ps1 file from the repo.

CIScripts/Build/BuildFunctions.ps1
```diff
@@ -256,6 +260,21 @@ function Invoke-AgentBuild {
...
 }
 
+function Copy-Dlls {
+    Param (
+       [Parameter(Mandatory = $true)] [string] $OutputPath,
+       [Parameter(Mandatory = $false)] [bool] $ReleaseMode = $false
+    )
+
+    $Job.Step("Copying dlls to $OutputPath", {
+        if (-not $ReleaseMode) {
```
m-kostrzewa: We're thinking that this `if` should be outside the `Copy-Dlls` function.

```
if(-not $ReleaseMode) {
 Copy-Dlls -OutputPath ......
}
function Copy-Dlls {
  params.
  job step
    cp x y 
}
```
Also, maybe we should rename the function to `Copy-VCDebugDlls`.

CIScripts/Build/BuildFunctions.ps1
```diff
@@ -196,7 +197,10 @@ function Invoke-ExtensionBuild {
         Copy-Item $utilsMSI $OutputPath -Recurse -Container
         Copy-Item $vRouterMSI $OutputPath -Recurse -Container
         Copy-Item $vRouterCert $OutputPath -Recurse -Container
-        Copy-Item $vTestPath "$OutputPath\utils\vtest" -Recurse -Container
+        New-Item -Type Directory -Path "$OutputPath\utils\vtest" | Out-Null
```
m-kostrzewa: In Build.ps1 we have all New-Item stuff. Let's move it there as well.

CIScripts/Deploy/Deployment.ps1
```diff
@@ -63,6 +63,9 @@ function Install-Artifacts {
...
+    Copy-Item -ToSession $Session -Path "dlls\*" -Destination "C:\Windows\System32\"
```
m-kostrzewa, line 67: I don't like that here we assume all the paths to artifacts.

In Build.ps1 we have all these
```

  $vRouterOutputDir = "output/vrouter"		  $vRouterOutputDir = "output/vrouter"
  $AgentOutputDir = "output/agent"		  $AgentOutputDir = "output/agent"
```
but now we don't use any variables like $vRouterOutputDir.

This is implicit global state.

IMHO we should move these paths to a config file in the future (not this PR) - will add this to backlog.

CIScripts/Build/BuildFunctions.ps1
```diff
@@ -196,7 +197,10 @@ function Invoke-ExtensionBuild {
...
+        Copy-Item "$vTestSrcPath\tests" "$OutputPath\utils\vtest" -Recurse -Filter "*.xml"
+        Copy-Item "$vTestSrcPath\*.ps1" "$OutputPath\utils\vtest"
+        Copy-Item "$vTestPath\*" "$OutputPath\utils\vtest" -Recurse -Container
```
m-kostrzewa: Maybe moving vtest and related xmls and scripts should be moved to a different Job.Step?

CIScripts/Build/BuildFunctions.ps1
```diff
@@ -178,6 +178,7 @@ function Invoke-ExtensionBuild {
     $vRouterMSI = "$vRouterRoot\extension\vRouter.msi"
     $vRouterCert = "$vRouterRoot\extension\vRouter.cer"
     $utilsMSI = "$vRouterRoot\utils\utils.msi"
+    $vTestSrcPath = "vrouter\utils\vtest\"
```
m-kostrzewa: K, if we're going with this, then we should rename $vRouterRoot to $vRouterBuildRoot or something. 

And then move $vTestSrcPath 2 lines below (so it's not mixed with build output paths). $vTestPath should also be renamed to $vTestBuilPath.

And maybe we should instead have $vTestSrcRoot and then two other vars:
$vTestScenariosPath (for xmls) and $vTestRunnerPath (for powershell script).

m-kostrzewa (changes requested):

m-kostrzewa:
reviewed-by MKos+DS

CIScripts/Build/BuildFunctions.ps1
krdln: Think of this `New-Item` as an implementation detail of the following `Copy-Item`s (Couldn't do it with a single `Copy-Item -Recurse`, because I needed two different filters, therefore split into separate `New-Item` + two `Copy-Item`s). Do you still think it should be moved to `Build.ps1`?

CIScripts/Build/BuildFunctions.ps1
krdln: If we're not going to copy any DLLs for release mode, then sure.

CIScripts/Test/TestRunner.ps1
krdln: Sorry, bad merge

CIScripts/Test/TestConfigurationUtils.ps1
krdln: I've changed that as per suggestion from @michal-clapinski, so let me check if the `New-ComntainerNetwork` works too.

krdln:
recheck no bug

CIScripts/Build/BuildFunctions.ps1
m-kostrzewa: I think the contract about artifact locations should be defined in one place. Why doesn't `-Container` option work here?

CIScripts/Deploy/Deployment.ps1
krdln: Testing and deployment assumes location of all artifacts, so the best way would be for this to be configured so that Build, Deploy and Test uses the same locations. But this is not in scope for this PR.

CIScripts/Deploy/Deployment.ps1
krdln: There are no other steps here, could you expand on what do you mean?

CIScripts/Test/Utils/CommonTestCode.ps1
krdln: I've just renamed `Install-Msi` to `Invoke-MsiExec`. I think splitting it into three functions is not necessary.

CIScripts/Test/TestConfigurationUtils.ps1
krdln: > Why are we switching to docker network create fron New-ContainerNetwork here?

I have a weird problem here – when using `New-ContainerNetwork` i have a problem with cleanup, namely `Get-ContainerNetwork | Remove-ContainerNetwork -Force` results in `Incorrect parameter` error. Not sure why's that happening though.

mclapinski: Yeah, I got it too. I had to do it twice (the second time it succeeded). No idea how it should be done in CI.

krdln: @michal-clapinski Do you think it's feasible to just this command twice, first time with `SilentlyContinue`?

mclapinski: That sounds awful but if you can't make it work any other way I think we may accept it.

CIScripts/Deploy/Deployment.ps1
m-kostrzewa: Every Write-Host in this file should instead be a Job.Step. This is outside of scope of this PR though.

CIScripts/Deploy/Deployment.ps1
m-kostrzewa: Agreed

CIScripts/Test/GetTestConfigurationJuni.ps1
```diff
@@ -0,0 +1,67 @@
+function Get-TestConfiguration {
+    [TestConfiguration] @{
+        ControllerIP = "10.7.0.54";
+        ControllerRestPort = 8082;
+        ControllerHostUsername = "ubuntu";
+        ControllerHostPassword = "ubuntu";
+        AdapterName = "Ethernet1";
+        VMSwitchName = "Layered Ethernet1";
```
m-kostrzewa, line 8: This used to be ` VMSwitchName = "Layered?Ethernet1";`, why the change to use a space instead of question mark?

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -159,6 +166,7 @@ function Disable-DockerDriver {
         Stop-Service docker | Out-Null
         Get-NetNat | Remove-NetNat -Confirm:$false
         Get-ContainerNetwork | Remove-ContainerNetwork -Force
+        Get-ContainerNetwork | Remove-ContainerNetwork -ErrorAction SilentlyContinue -Force
```
m-kostrzewa: Could you explain this change? Also, why is it almost a duplicate of line 168?

m-kostrzewa (dismissed):
Thanks for using PSSessionT throughtout the codebase.

Checked alone, without @sodar (he's gone home)

CIScripts/Test/GetTestConfigurationJuni.ps1
krdln: The question mark was introduced https://github.com/codilime/contrail-windows-ci/commit/0e52ce06dc0c9f0e369f2bca2414eabba7fd9ed3 but it was then reverted. So the question mark is no longer needed

CIScripts/Test/TestConfigurationUtils.ps1
krdln: These lines should be in reversed order. Could you quickly commit the reversal and let' see what happens? 

For some reason, this line fails in CI, I didn't have time to properly investigate it, so I made this line just run twice as a temporary hack. This should be fixed properly though, either in this or follow up pr.

m-kostrzewa (approved):

krdln:
The workaround for error in `Remove-ContainerNetwork` should be cleaned up in follow-up PR.

PR #86, opened 2018-02-09
Zuul marks projects not handled by WinCI as 'good'

jacek-izykowski:


ansible/roles/zuul/files/layout.yaml
```diff
@@ -85,3 +89,153 @@ projects:
   - name: Juniper/contrail-common
     template:
       - name: ci-contrail-windows-template
+
+# AUTOMATICALLY GENERATED NOOPs BELOW - DO NOT MODIFY THESE LINES MANUALLY
+
+  - name: All-Users
```
kaweue: Is this really required ?

kaweue (dismissed):
KR + PM

ansible/roles/zuul/files/layout.yaml
jacek-izykowski: It's not. But it's also not harmful though.

This list has been automatically generated. Do you suggest to add this specific name as a special case for generating script?

jablonskim (approved):

katrybacka (approved):

PR #92, opened 2018-02-13
Add testbed role and clean up waiting for dhcp

krdln:
cc @michal-clapinski


ansible/roles/vmware-template/tasks/create.yml
```diff
@@ -57,6 +53,10 @@
     datacenter: "{{ datacenter_name }}"
 
     uuid: "{{ new_vm.instance.hw_product_uuid }}"
+  retries: 60
+  delay: 15
+  until: >
+      vm.instance.hw_eth0.ipaddresses | ipv4 | reject("match", "169.254.*") | list
```
sodar, line 59: Are you that this condition will not fail if `vm.instance.hw_eth0.ipaddresses` can be null? When VMware does not see any addresses `vm.instance.hw_eth0.ipaddresses` can be null, not an empty list.

ansible/roles/vmware-template/tasks/deploy.yml
```diff
@@ -51,19 +46,15 @@
     datacenter: "{{ datacenter_name }}"
 
     uuid: "{{ new_vm.instance.hw_product_uuid }}"
+  retries: 60
+  delay: 15
+  until: >
+      vm.instance.hw_eth0.ipaddresses | ipv4 | reject("match", "169.254.*") | list
```
sodar, line 52: Same as above comment.

ansible/vmware-deploy-testenv.yml
```diff
@@ -37,3 +37,11 @@
   roles:
   - role: vmware-vm
     action: deploy
+
+- hosts: testbed
```
sodar: When we will use controller in tests we should change destination `hosts` from `testbed` to `testenv`. Maybe we should add this to backlog and put a `TODO` comment here?

sodar (changes requested):

ansible/roles/vmware-template/tasks/create.yml
krdln: I've just checked and if the field is null, the expression correctly evaluates to false. (if the `ipaddresses` field were not present, this would be a problem thought)

<details>
  <summary>Example</summary>
   role:

```yaml
- name: set
  set_fact:
    dict: {'aaa': null, 'bbb': 'null'}

- name: debugdict
  debug:
     var: dict

- name: foobar
  register: xxx
  win_shell: Write-Host bar
  retries: 10
  delay: 1
  until: dict.aaa | ipv4 | reject("match", "aaa") | list
```

output:

```
TASK [Gathering Facts] ************************************************************************************************************************************************************************
ok: [10.84.12.185]

TASK [foo : set] ******************************************************************************************************************************************************************************
ok: [10.84.12.185]

TASK [foo : debugdict] ************************************************************************************************************************************************************************
ok: [10.84.12.185] => {
    "dict": {
        "aaa": null, 
        "bbb": "null"
    }
}

TASK [foo : foobar] ***************************************************************************************************************************************************************************
FAILED - RETRYING: foobar (10 retries left).
FAILED - RETRYING: foobar (9 retries left).
```
</details>

krdln:
Rebased

sodar (approved):
Looks good to me. Reviewed with @jacek-izykowski.

PR #95, opened 2018-02-15
System independent jenkins partition role

jablonskim:


ansible/roles/jenkins-partition/tasks/linux.yml
```diff
@@ -0,0 +1,21 @@
...
+  become: true
+
+- name: Create filesystem
+  filesystem:
+    dev: /dev/sdb1
+    fstype: ext4
+  become: true
+
+- name: Mount partition
+  mount:
```
sodar, line 17: How is this working without `become: true`?

sodar (changes requested):

sodar (approved):

PR #99, opened 2018-02-16
Fix timestamps format in logs

krdln:
[Example](http://logs.opencontrail.org/winci/35638fc7e6f44769bf13fa74841f293e/log.txt.gz) (timestamps are now clickable!)

Also, added compression of logs.


jacek-izykowski (dismissed):
KR, JI

Jenkinsfile
```diff
@@ -162,10 +161,8 @@ pipeline {
                         // unstash "buildLogs"
                         // TODO correct flags for rsync
                         sh "ssh ${LOG_SERVER_USER}@${LOG_SERVER} \"mkdir -p ${LOG_ROOT_DIR}/${BUILD_SPECIFIC_DIR}\""
-                        // The timestamps are not stored on disk as raw text, but in some encoded form,
-                        // so the easiest way to decode them is to use http path provided by the timestamper plugin.
-                        sh "curl --silent 'http://localhost:8080/$RAW_LOG_PATH' --output clean_log.txt"
-                        sh "rsync clean_log.txt ${LOG_SERVER_USER}@${LOG_SERVER}:${REMOTE_DST_FILE}"
+                        obtainLogFile env.JOB_NAME, env.BUILD_ID, "clean_log.txt.gz"
```
katrybacka: I see that we use ```()``` around arguments when we call groovy scripts in whole Jenkinsfile, what do you think about sticking to this convention here?

katrybacka: I think that we could use ``` clean_log.txt.gz ``` as a variable since we use it twice here and const string looks a little strange for me due to possible future changes of name of file or something.

katrybacka (changes requested):

Jenkinsfile
krdln: > what do you think about sticking to this convention here?

Do we have such a convention, or did it just happen an accident? Personally, I prefer omitting `()`.

> clean_log.txt.gz as a variable

Sure, will change.

katrybacka: As far as I know we use ```()``` everytime we call a script, so I prefer to use one convention. We can discuss it with other members of our team to use one option, what do you think about that?

Jenkinsfile
```diff
@@ -161,8 +161,9 @@ pipeline {
                         // unstash "buildLogs"
                         // TODO correct flags for rsync
                         sh "ssh ${LOG_SERVER_USER}@${LOG_SERVER} \"mkdir -p ${LOG_ROOT_DIR}/${BUILD_SPECIFIC_DIR}\""
-                        obtainLogFile env.JOB_NAME, env.BUILD_ID, "clean_log.txt.gz"
-                        sh "rsync clean_log.txt.gz ${LOG_SERVER_USER}@${LOG_SERVER}:${REMOTE_DST_FILE}"
+                        def tmpLogFilename = "clean_log.txt.gz"
+                        obtainLogFile env.JOB_NAME, env.BUILD_ID, tmpLogFilename
```
katrybacka: I see ```()```  everywhere a groovy script is called and I think we should choose one option, we can discuss with other members of our team and decide which convention to use. What do you think about that?

katrybacka (changes requested):

krdln:
Works after merge: [Submitted log example](http://logs.opencontrail.org/winci/35638fc7e6f44769bf13fa74841f293e/log.txt.gz), [Jenkins link](http://10.84.12.26:8080/job/WinContrail/job/winci-server2016-devel/175/)

katrybacka (approved):

PR #101, opened 2018-02-16
Add quotes to all python invocations

krdln:


ansible/roles/vmware-template/tasks/deploy.yml
```diff
@@ -45,7 +45,7 @@
 
     datacenter: "{{ datacenter_name }}"
 
-    uuid: "{{ new_vm.instance.hw_product_uuid }}"
+    id: "{{ new_vm.instance.hw_product_uuid }}"
```
sodar: Are you certain that `vmware_guest_facts` accepts `id` as a parameter? According to https://github.com/ansible/ansible/blob/v2.4.0.0-1/lib/ansible/modules/cloud/vmware/vmware_guest_facts.py#L132 it accepts `uuid`, not `id`.

vars/unlockNetwork.groovy
```diff
@@ -10,6 +10,6 @@ def call(String networkName) {
             '--network-name', networkName,
         ]
 
-        sh cmd.join(' ')
+        sh cmd.collect{ arg -> "'" + arg + "'" }.join(' ')
```
sodar: Maybe we should move this fragment into separate function? It might be useful in the future.

sodar (changes requested):

ansible/roles/vmware-template/tasks/deploy.yml
krdln: You're right. I haven't intended to make this change.

vars/unlockNetwork.groovy
krdln: I've extracted it to `shellCommand` function. Also added escaping according to [this SO answer](https://unix.stackexchange.com/questions/187651/how-to-echo-single-quote-when-using-single-quote-to-wrap-special-characters-in/187654#187654).

sodar (dismissed):

krdln:
@sodar I've added a test for the `shellCommand` helper. What do you think about this approach?

Also, I've let the [gather-build-stats use this helper, and it looks like it works](http://10.84.12.26:8080/job/WinContrail/job/gather-build-stats/30/)

sodar (approved):

PR #105, opened 2018-02-19
Strip credentials from logs in DemoEnv

jablonskim:


krdln:
Could you post some links to *why* this works?

(also, please add a link to example log)

krdln:
In https://github.com/codilime/contrail-windows-ci/pull/104 the `VC = credentials('vcenter')` line was moved out from toplevel `environment`. But from what is see from [its log](http://10.84.12.26:8080/job/WinContrail/job/winci-server2016-contrail-windows-ci/437/console) the `****` are still present there. Could you explain this behaviour?

Also, please add comments near the assignments explaining why it shouldn't be moved.

jablonskim:
The `VC = credentials('vcenter')` line works only in scope where it has been defined. In https://github.com/codilime/contrail-windows-ci/pull/104 the `VC_USR` and `VC_PSW` are used only inside `Cleanup-Provision-Deploy-Test`, so they are stripped correctly. In case of this PR, credentials are also used inside `Run ansible stage` so it should be moved to outer scope.

krdln:
So the problem was "leaking" credentials from one stage to another using a file, right? What do you think about adding a following line
```groovy
assert VC // make sure credentials are seen by Jenkins, so they can be censored
```

before
```groovy
ansiblePlaybook extras: '-e @vm.vars -e @common.vars' ...
```

(both in these files and in main `Jenkinsfile`)

jablonskim:
I discussed this idea with PM and we think that there is no need to add this `assert` there:

1. I have already added comment near the assignment.
2. Assert in my opinion shouldn't be used in production code (it is intended to use more frequently with test code).
3. In the future, for example, someone may add another stage - in this case comment near assignment is safer than an `assert` in specific stage.

krdln (approved):

krdln (changes requested):
Please add a comment to the main `Jenkinsfile` too.

krdln (approved):

PR #107, opened 2018-02-19
Fix "double Remove-ContainerNetwork" workaround

krdln:
Tries to fix workaround introduced in https://github.com/codilime/contrail-windows-ci/pull/82

When using:
* `New-ContainerNetwork`
* `Remove-ContainerNetwork`

the latter fails with "Incorrect parameter"
and `Remove-ConatinerNetwork` has to be redone.
This is how `Enable-VRouterExtension` works currently.

Similar mode of failure occurs with:
* `docker network create`
* `docker network rm`

Somehow, the clean path is:
* `docker network create`
* `docker network prune` – this actually causes the same HNS error as `rm` (*The parameter is incorrect*), but `prune` exits cleanly regardless.
* `Remove-ContainerNetwork`

Which is what this commit uses

---

See https://github.com/docker/for-win/issues/1239

### DIFF
--- a/CIScripts/Test/GetTestConfigurationJuni.ps1
+++ b/CIScripts/Test/GetTestConfigurationJuni.ps1
@@ -5,7 +5,11 @@ function Get-TestConfiguration {
         ControllerHostUsername = "ubuntu";
         ControllerHostPassword = "ubuntu";
         AdapterName = "Ethernet1";
-        VMSwitchName = "Layered Ethernet1";
+
+        # When creating network using `docker network create`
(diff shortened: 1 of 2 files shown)


krdln:
cc @m-kostrzewa @jablonskim

krdln:
Actually, I'm not sure which workaround is worse – we either do "double-Remove" explicitly or we do "double-Remove" using `prune` as the first `Remove`.

krdln:
Let's not do this, as it is only hiding a bug.

PR #109, opened 2018-02-20
vmware: correctly handle std networks in provision_vm.py

sodar:


sodar:
recheck no bug

kaweue (approved):
MJ + PM

PR #110, opened 2018-02-20
JW-1433 : Adding controller

kaweue:
Adding roles contrail-ansible-deployer, kolla-provision-docker and controller


kaweue:
@katrybacka FYI

ansible/roles/controller/tasks/main.yml
```diff
@@ -0,0 +1,37 @@
...
+- name: Install build dependencies
+  package:
+    name: "{{ item }}"
+    state: latest
+  become: true
+  with_items:
+    - python2-requests
+    - git
+
+- include_role:
```
katrybacka: Can we use list of roles here instead of using ```include role``` twice?

katrybacka (dismissed):

ansible/roles/controller/tasks/main.yml
jablonskim: +1

jablonskim (changes requested):

katrybacka (approved):
MJ + KR

kaweue:
recheck no bug

jablonskim (approved):

kaweue:
recheck no bug

PR #112, opened 2018-02-20
Fix COPY_DISABLED_ARTIFACTS logic

krdln:
Copy only nonexistent items.

Splitted in two commits for clearer diff.

Note to self: Turn on COPY_DISABLED_ARTIFACTS by default in CI after merging.


krdln:
[Works](http://10.84.12.26:8080/job/WinContrail/job/winci-server2016-contrail-windows-ci/512/console)

CIScripts/BuildStage.ps1
```diff
@@ -9,6 +9,10 @@ $Credentials = New-Object PSCredentialT ($Username, $Password)
 $NothingToBuild = $Env:COMPONENTS_TO_BUILD -eq "None"
 $CopyDisabledArtifacts = Test-Path Env:COPY_DISABLED_ARTIFACTS
 
+if (-not $NothingToBuild) {
+    & $PSScriptRoot\Build.ps1
+}
+
 if ($NothingToBuild -or $CopyDisabledArtifacts) {
```
m-kostrzewa, line 16: Maybe this should be an elseif then?

CIScripts/BuildStage.ps1
```diff
@@ -23,13 +27,21 @@ if ($NothingToBuild -or $CopyDisabledArtifacts) {
...
 
-    $Job.Done()
-}
+    foreach ($Item in Get-ChildItem "${DiskName}:\") {
+        $OutputItem = "$OutputRootDirectory\$($Item.Name)"
 
-if (-not $NothingToBuild) {
-    & $PSScriptRoot\Build.ps1
+        if ((Test-Path $OutputItem) -And (Get-ChildItem $OutputItem | Measure-Object).Count -ne 0) {
+            # Skip existing files and non-empty directories
```
m-kostrzewa: This comment wouldn't be necessary if we named the elements of `if` statement. Something like

```
$FileAlreadyExists = Test-Path $OutputItem
$IsNonEmptyDirectory = (Get-ChildItem $OutputItem | Measure-Object).Count -ne 0
if($FileAlreadyExists -Or $IsNonEmptyDirectory)
{ continue }
```

Also, shouldn't there be an -Or instead of -And?

m-kostrzewa (changes requested):

CIScripts/BuildStage.ps1
krdln: We can't extract the conditions to variables because lhs serves as guard for rhs.

And yes, `and` is the correct behaviour here. With `or` we'd skip empty directories too.

Anyway, I have an idea how to simplify this condition, so stay tuned.

CIScripts/BuildStage.ps1
krdln: These ifs are not disjoint, so I don't think it'd be possible.

m-kostrzewa: How about

if(-not $NothingToBuild) {
..
} elseif ($CopyDisabledArtifacts) {
...
}

CIScripts/BuildStage.ps1
m-kostrzewa: That's what I suspected but didn't really read into it since I decided it's not readable enough as it is. I think we should avoid if statements that do anything more than check a condition if possible.

CIScripts/BuildStage.ps1
krdln: We can't use `else` because we must to be able to enter both first and second if (as I said before, they're *not disjoint*)

m-kostrzewa: nvm I'm dumb

m-kostrzewa (approved):

krdln:
recheck no bug

krdln:
recheck no bug

PR #117, opened 2018-02-22
Add stage for running tests for python scripts

krdln:


krdln:
recheck no bug

krdln:
recheck no bug

Jenkinsfile
```diff
@@ -61,6 +62,20 @@ pipeline {
             }
         }
 
+        stage('Linux test') {
```
jacek-izykowski: MJ - there was some issue with stage names containing spaces. We were removing them recently.

Jenkinsfile
```diff
@@ -61,6 +62,20 @@ pipeline {
...
+            when { expression { env.ghprbPullId } }
```
jacek-izykowski: This stage is going to be executed for all repositories from GitHub, which is not necessary (including https://github.com/Juniper/contrail-windows).

jacek-izykowski (changes requested):

Jenkinsfile
krdln: The goal was just to make gerrit logs cleaner and don't let random people wonder "why are linux tests running in windows CI". I don't think `contrail-windows` would be a problem here.

jacek-izykowski (approved):

jablonskim (approved):

PR #119, opened 2018-02-22
Exclude git from SourceCode stash

krdln:


krdln:
Git dirs are excluded by default.

PR #120, opened 2018-02-22
Don't look at Components_to_build in deploy

krdln:


CIScripts/Deploy/Deployment.ps1
```diff
@@ -17,15 +17,18 @@ function Install-Artifacts {
     Param ([Parameter(Mandatory = $true)] [PSSessionT] $Session,
            [Parameter(Mandatory = $true)] [string] $ArtifactsDir)
 
+    function Test-NonemptyDir {
```
m-kostrzewa, line 20: IMHO this should be moved to Common/ dir

krdln: I don't feel like creating a separate file for a one-liner function. Is there any file I can add it to?

jablonskim (approved):
MJ & MC

CIScripts/Deploy/Deployment.ps1
m-kostrzewa: I guess we'll see if some other part of the codebase requires this function.

PR #123, opened 2018-02-22
monitoring: trigger 'gather-build-stats' job

sodar:


Jenkinsfile
```diff
@@ -191,6 +191,11 @@ pipeline {
                         deleteDir()
                     }
                 }
+
+                build job: 'WinContrail/gather-build-stats',
```
jablonskim: Please add `wait: false`.
There will be a race condition. It should be fixed in new PR.

jablonskim (changes requested):
MJ & JI

jacek-izykowski (changes requested):
gather-build-stats always receives "IN PROGRESS".

jacek-izykowski (dismissed):
There's race condition now, but as we agreed - that's going to be handled in next PR.

jablonskim (approved):

jacek-izykowski (approved):

